PR #19, opened 2015-08-07
Update bootstrap and buildout

gforcada:
Just a testing PR to double check that the errors I have locally are only my own problems and not general ones.


src/z3c/dependencychecker/tests/sample1/setup.py_in
```diff
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 
 
 setup(name='sample1',
-      version='0.2dev',
+      version='0.2.dev0',
```
gforcada: The usual ```UserWarning: version number normalization...`` from setuptools was getting into some output and preventing a file being found.

PR #21, opened 2015-08-08
Use newer travis infra

gforcada:


gforcada:
Moved to #19 so only one PR needs to be reviewed.

PR #26, opened 2015-08-09
Add FTI support

gforcada:
Plone add-ons may install new content types,
which are defined in XML files where some dependencies maybe hidden.

### DIFF
--- a/src/z3c/dependencychecker/dependencychecker.py
+++ b/src/z3c/dependencychecker/dependencychecker.py
@@ -84,6 +84,44 @@ for=         #
 ['\"]        # Single or double quote.
 """, re.VERBOSE)
 
+FTI_BEHAVIOR_PATTERN = re.compile(r"""
+\s           # Whitespace.
+<element     #
+\s+          #
+value=       #
+['\"]        # Single or double quote.
+(?P<import>  # Start of 'import' variable.
+\S+          # Non-whitespace string.
+)            # End of 'import' variable.
+['\"]        # Single or double quote.
+""", re.VERBOSE)
+
+FTI_SCHEMA_PATTERN = re.compile(r"""
+\s           # Whitespace.
+name=        #
+['\"]        # Single or double quote.
+schema       #
+['\"]        # Single or double quote.
+>            # End of opening tag
+(?P<import>  # Start of 'import' variable.
+\S+          # Non-whitespace string.
+)            # End of 'import' variable.
+<            # Start of closing tag
+""", re.VERBOSE)
+
+FTI_KLASS_PATTERN = re.compile(r"""
+\s           # Whitespace.
+name=        #
+['\"]        # Single or double quote.
+klass        #
+['\"]        # Single or double quote.
+>            # End of opening tag
+(?P<import>  # Start of 'import' variable.
+\S+          # Non-whitespace string.
+)            # End of 'import' variable.
+<            # Start of closing tag
+""", re.VERBOSE)
+
 DOCTEST_IMPORT = re.compile(r"""
 ^            # From start of line
 \s+          # Whitespace.
@@ -343,6 +381,33 @@ def includes_from_zcml(path):
     return modules, test_modules
 
 
+def includes_from_plone_fti(path):
+    modules = []
+    test_modules = []
+    for path, dirs, files in os.walk(path):
+        for fti in [os.path.abspath(os.path.join(path, filename))
+                    for filename in files
+                    if fnmatch.fnmatch(filename, '*.xml')]:
+            contents = open(fti).read()
+            # check that we are on a Dexterity FTI and
+            # not any other random XML file
+            if contents.find('meta_type="Dexterity FTI"') == -1:
+                continue
+
+            found = [module for module in
+                     re.findall(FTI_BEHAVIOR_PATTERN, contents)
+                     if module.find('.') != -1]
+            found += [module for module in
+                      re.findall(FTI_SCHEMA_PATTERN, contents)]
+            found += [module for module in
+                      re.findall(FTI_KLASS_PATTERN, contents)]
+            if 'tests' in fti:
+                test_modules += found
+            else:
+                modules += found
+    return modules, test_modules
+
+
 def includes_from_generic_setup_metadata(path):
     modules = []
     test_modules = []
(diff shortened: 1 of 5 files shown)


src/z3c/dependencychecker/dependencychecker.py
```diff
@@ -84,6 +84,57 @@
 ['\"]        # Single or double quote.
 """, re.VERBOSE)
 
+FTI_BEHAVIOR_PATTERN = re.compile(r"""
+\s           # Whitespace.
+<element     #
```
jone: Wouldn't this expression match non-behavior element tags too?
```xml
 <property name="view_methods">
  <element value="document_view"/>
 </property>
```
https://github.com/plone/plone.app.contenttypes/blob/master/plone/app/contenttypes/profiles/default/types/Document.xml

src/z3c/dependencychecker/dependencychecker.py
```diff
@@ -84,6 +84,57 @@
...
+
+FTI_SCHEMA_PATTERN = re.compile(r"""
+\s           # Whitespace.
+name=        #
+['\"]        # Single or double quote.
+schema       #
+['\"]        # Single or double quote.
+>            # End of opening tag
+(?P<import>  # Start of 'import' variable.
+\S+          # Non-whitespace string.
```
jone: I personally prefer to use explicit lookaheads in such situations:
`[^<]+` would match all characters but not a `<`; `\S` would match `<` too, which makes it less specific: if the expression is too greed we would probably have too much in the `import` group.

For instance when there is a typo (two `<` characters), it wouldn't match correctly:
```python
>>> import re
>>> re.search(r'>(\S+)<', '<foo>bar</foo>').group(1)
'bar'
>>> re.search(r'>(\S+)<', '<foo>bar<</foo>').group(1)
'bar<'
```

although one could argument that then the XML would be invalid and therefore it is okay when the result is wrong.
Just my thoughts :wink:

src/z3c/dependencychecker/dependencychecker.py
```diff
@@ -343,6 +394,39 @@ def includes_from_zcml(path):
...
+                    if fnmatch.fnmatch(filename, '*.xml')]:
+            contents = open(fti).read()
+            # check that we are on a Dexterity FTI and
+            # not any other random XML file
+            if contents.find('meta_type="Dexterity FTI"') == -1:
+                continue
+
+            found = [module for module in
+                     re.findall(FTI_BEHAVIOR_PATTERN, contents)
+                     if not module.startswith('.') and module.find('.') != -1]
```
jone: Why are we filtering matches with leading dots here? It shouldn't be possible to have relative dotted names here, it always must start with a dot in order to work properly, I think.

src/z3c/dependencychecker/dependencychecker.py
```diff
@@ -343,6 +394,39 @@ def includes_from_zcml(path):
...
+
+            found += [module for module in
+                      re.findall(FTI_SCHEMA_PATTERN, contents)
+                      if not module.startswith('.')]
+            found += [module for module in
+                      re.findall(FTI_KLASS_PATTERN, contents)
+                      if not module.startswith('.')]
+            found += [module for module in
+                      re.findall(FTI_PERMISSION_PATTERN, contents)
```
jone: Permissions are not real dotted names, it is just the name of the `IPermission` utility and it could be any string.
For example the permission [`plone.app.contenttypes.addDocument`](https://github.com/plone/plone.app.contenttypes/blob/master/plone/app/contenttypes/profiles/default/types/Document.xml#L20) cannot be imported, since it is just a name.
What is the idea behind including this?

src/z3c/dependencychecker/dependencychecker.py
gforcada: That's why on the list comprehension it filters any match that does not have a dot in it, we could encode that into the regex of course.

I was just lazy and using the same logic as on the surrounding methods (I basically copy&pasted&adapted ZCML detection code).

src/z3c/dependencychecker/dependencychecker.py
gforcada: My assumption here is that the package is in a sane state, if the XML is broken, maybe is actually even a good idea to suggest broken missing imports so that they can grep for those, don't you think?

src/z3c/dependencychecker/dependencychecker.py
gforcada: Then we could remove that, thanks for the tip, I basically copy&pasted&adapted ZCML, it was a late night hack, so expect (as you already have seen) lots of room for improvement. Thanks for the review btw.

src/z3c/dependencychecker/dependencychecker.py
gforcada: I basically skimmed over an FTI looked what seemed to be a dotted name and put it there, if you are sure that they are meaningless in regards of dependencies I will remove them, thanks for pointing it out!

gforcada:
@jone thanks for the review I hope to fix the issues raised by you within some days.

src/z3c/dependencychecker/dependencychecker.py
reinout: You can pretty much safely asume that the package is in a sane state. Once you start using a dependencychecker to clean up your dependency graph, you're pretty much guaranteed to have well-behaved packages.

So it doesn't need to be made more elaborate or fail-safe than now, I think.

reinout:
@gforcada: ping me when the pull request is ready according to you, ok?

src/z3c/dependencychecker/dependencychecker.py
jone: :+1: I'm perfectly okay with leaving it like it is. I just noticed it and realized that I'd have done it differently, but I think it is perfectly fine as it is too. :wink:

src/z3c/dependencychecker/dependencychecker.py
jone: I think the problem is that we don't know in which context the `<element>` tag appears.
Regex is not really useful for parsing context-aware things such as XML, at least when the context gets relevant as it is the case here (seeh the [Chomsky hierarchy](https://en.wikipedia.org/wiki/Chomsky_hierarchy)).

Maybe it would be better to use a XML parser, such as lxml or minidom, what do you think?

reinout: If possible, I'd just do it with a regex.

Reason: this way it is the same kind of code as the other checks.

A reason to use an xml parser: a *specific* example in a public plone package for which the regex would fail and for which a small regex fix is not possible. The string `maybe` isn't a good enough example :-)

gforcada:
@reinout if @jone does not have any more concern I already fixed the issues he raised. So ready to be merged and released :)

reinout:
I released 1.12. Can you check if it works for you?

Thanks for the work!

gforcada:
Sorry for the late feedback...

Yes it works perfectly and I already found some more places where dependencies could be found. I will cook a PR later.

reinout:
Ok, thanks!

PR #27, opened 2015-08-09
Add plone.recipe.codeanalysis

gforcada:
Add static code analysis tools that run on git pre-commit hook as
well as anytime with ./bin/code-analysis

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -6,6 +6,7 @@ parts =
     test
     omelette
     console_scripts
+    code-analysis
 develop =
     .
 eggs =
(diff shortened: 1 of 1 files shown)


reinout:
I am a big fan of pep8 and pyflakes. Normally I would not want it in a pre-commit hook, but a test cannot hurt :-) I didn't know that code analysis recipe existed. Is that pretty much standard in plone packages at the moment?

gforcada:
Yes, at least we are trying to push it as  the standard style guide for Plone.

reinout:
I'm starting to like it. If you run it on master, the output should be completely clean now.

I *did* switch off the import check as I'm used to the pep sorting instead of the plone sorting. Though I must say that the plone sorting *does* look nice. And with the upcoming isort integration (https://github.com/plone/plone.recipe.codeanalysis/issues/148)...

PR #28, opened 2015-08-28
Handle platform specific naming conventions

gforcada:
time is a module from Python standard library:
https://docs.python.org/2/library/time.html

Still z3c.dependencychecker thinks it is an external distribution
that must be declared as a dependency.

### DIFF
--- a/src/z3c/dependencychecker/dependencychecker.py
+++ b/src/z3c/dependencychecker/dependencychecker.py
@@ -328,6 +328,9 @@ def _detect_modules(sample_module):
     for stdlib_file in stdlib_files:
         module, extension = os.path.splitext(stdlib_file)
         if extension == stdlib_extension:
+            if module.endswith('module'):
+                # See http://stackoverflow.com/q/6319379/27401
+                module = module[:-len('module')]
             modules.append(module)
     if 'py' in stdlib_extension:
         # Also check directories with __init__.py* in them.


gforcada:
While using z3c.dependencychecker on a distribution that is using ``from time import time`` it reported ``time`` as a missing dependency.

I'm not sure where to start digging for the error here, any pointer?

reinout:
Eh, I assume the test was supposed to fail? But it succeeds. Or am I missing something?

reinout:
In dependencychecker.py, look at the `stdlib_modules()` function. 

It uses `_detect_modules()` to find `.so` and `.py` files and so on based on one sample module. `stdlib_modules()` calls it with "os" and "datetime". That combination finds almost everything. At the end, a couple of extra libraries are added.

So... you could add "time" by hand there. And otherwise find out why 'time' isn't found. Which OS/version do you have?

gforcada:
@reinout thanks for the pointers, it does seem indeed that there's something wrong on my local setup as travis checks did pass, and it does not locally here.

I'm using Linux (Fedora 20) and python 2.7.10

I will look at what you suggested and report back, thanks again!

gforcada:
It seems to be related to 64 bits.

On ``stdlib_modules`` function both ``py_module`` and ``dynload_module`` point to my virtualenv ``lib64`` folder where you find quite a few modules (no idea why) that end with ``..module.so`` like ``timemodule.so`` or ``itertoolsmodule.so`` ...

I'm trying to gather information about why that renaming is happening at all but so far nothing pops up...

reinout:
See http://stackoverflow.com/questions/6319379/python-shared-object-module-naming-convention

So... looks like:

    for stdlib_file in stdlib_files:
        module, extension = os.path.splitext(stdlib_file)
        if extension == stdlib_extension:
            modules.append(module)

should become something like:

    for stdlib_file in stdlib_files:
        module, extension = os.path.splitext(stdlib_file)
        if extension == stdlib_extension:
            if module.endswith('module'):
                module = module[:-len('module')]
            modules.append(module)

gforcada:
wow, thanks for the link, my search skills are lacking behind it seem :)

I will update the code of this PR then.

gforcada:
And I can confirm that it works.

gforcada:
@reinout thanks! :beers:

PR #29, opened 2015-08-28
Only alphanumeric characters on imports

gforcada:
Using \S+ on a regular expression matches 'transaction;' as ';' is a
non-space character.

Changing the 'S' by a 'w' (alphanumeric characters) fixes it.

### DIFF
--- a/src/z3c/dependencychecker/USAGE.rst
+++ b/src/z3c/dependencychecker/USAGE.rst
@@ -60,6 +60,7 @@ script would:
          plone.dexterity.browser.views.ContentTypeView
          plone.dexterity.interfaces.IContentType
          reinout.hurray
+         transaction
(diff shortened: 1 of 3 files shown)


reinout:
The fix breaks all imports with a dot in it like "import z3c.testsetup". Like the travis tests show: `zope.testbrowser` is now an unneeded test requirement and `zope` is missing.

Can you run the tests locally and try if `\w\.` instead of only `\w` works? That ought to allow dots in the name.

gforcada:
Sorry my mistake not to run tests locally first!

Now it should work.

reinout:
Thanks, looks like a good addition.

reinout:
See https://github.com/reinout/z3c.dependencychecker/commit/bd6e9526131da3b3dcfd2069d5a5d76c3237cceb :-)

PR #30, opened 2015-08-28
Look for imports on ZCML subscribers

gforcada:
ZCML subscribers have the 'handler' attribute where an import can be found.


reinout:
Thanks, looks good!

PR #34, opened 2015-08-28
Structure cleanup: moved src/z3c/ to just z3c/

reinout:


reinout:
Doing it directly. **IF** there are any pull requests or branches that don't have this change yet, I *think* git will do the right thing. It shouldn't inconvenience anyone too greatly.

Note that you probably will have to re-run buildout!

Issue #23, opened 2015-08-08
Marks as unneeded a zope package


gforcada:
Having ``zope.schema`` on ``setup.py`` and using it like this:

``from zope import schema``

makes ``z3c.dependencychecker`` think that ``zope.schema`` is not needed as ``Zope`` is already a package? My current guess.

Not sure how to handle this nor where to start looking for.

jone:
Is `runpy` available at the python you are using? Can you `import runpy`?

Actually it should work like this (when `runpy` is available):
The runpy-loader is asked for the file path of a dottedname, e.g.:
```python
>>> import runpy
>>> loader = runpy.get_loader('zope.schema')
>>> loader.get_filename()
'/Users/jone/projects/python/cache/eggs/zope.schema-4.2.2-py2.7.egg/zope/schema/__init__.py'
```
Then the `PKG-INFO`-file is searched (is located in an egg-info directory of any parent directory of this path).
In the `PKG-INFO` file there is a line with the package name:
```
Name: zope.schema
```

When runpy is not available, the dottedname is used without further lookup / conversion.

--

But I have also a strange behavior I cannot currently explain:
I have a `zope.schema` import in the code and a `zope.schema` dependency in the `setup.py` but the dependencychecker still lists it in `Missing requirements`, although I have declared the dependency.

gforcada:
I will check, thanks for the tip

gforcada:
Yes I do have ``runpy`` available.

For me it currently tells me that that is an unneeded dependency...

reinout:
Can you run dependencychecker with the `-v` option and post the results? Perhaps that'll help in figure out what goes wrong.

gforcada:
I will do and report back.

gforcada:
Aha! I think I have a better idea, still need to look at the code though...

The imports look like this:
```
from z3c.form import form
from z3c.form import interfaces
from zope import interface
from zope import schema
```

And the debug information reports it as:

```
DEBUG: Modules found in /home/gil/foo.py (test-only: False):
    ['z3c.form.form', 'zope.interface']
```

So as they share the same root (zope) it just ignores any further import from the same root (just like with z3c.form).

If I reverse the order (first zope.schema then zope.interface) it shows the opposite as expected.

I will try to cook a PR for it.

reinout:
Maybe #16 is related?

gforcada:
[...]

PR #36, opened 2015-08-28
Handle imports with the same root on multiple lines

gforcada:
Up until this commit the second import was completely ignored:

```
from zope import interface, something, bananas
from zope import component, whatnot
```

With this commit ``zope.component`` and ``zope.whatnot`` will be included
as dependencies.

Fixes https://github.com/reinout/z3c.dependencychecker/issues/23

### DIFF
--- a/z3c/dependencychecker/USAGE.rst
+++ b/z3c/dependencychecker/USAGE.rst
@@ -53,6 +53,7 @@ script would:
          plone.random2.content.MyType
          some_django_app
          something.origname
+         zope.exceptions
          zope.interface
     <BLANKLINE>
     Missing test requirements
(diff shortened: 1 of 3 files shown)


reinout:
Wow! This is quite a big change! Thanks for the fix.

I don't completely understand *everything* that is happening in the pull request, but that's mostly because the original "importchecker" code that I copied from somewhere deep in zope isn't that clearly written. I'll probably go through the entire module later on to try and clean it up a bit.

Anyway, thanks a lot!

I guess this would be a good time to do a fresh release? A release is just a minute of work so if there's another change afterwards I can easily do another.

gforcada:
Thanks for merging!

I was thinking the same (cleaning and modernizing a bit importchecker) but is quite a major effort... I will see if I have time at some point.

But a release would be super nice to have, and so far I haven't found anything else that's failing on the distributions that I checked.

reinout:
Well, I will have to do some cleaning/modernizing anyway when I want to get this package ready for python 3 :-)

I'm first going to do some more work on the django integration, as that's where I need it for myself.

PR #39, opened 2015-09-01
Fix when multiple egg-info files are found in a directory

reinout:
See changelog


reinout:
importchecker isn't really tested at the moment, so adding a test is too much work right now.

And the runpy stuff will probably be replaced by AST, so...

PR #43, opened 2016-06-11
Minor cleanup

mauritsvanrees:
Code analysis fixes and small efficiency fixes.

### DIFF
--- a/z3c/dependencychecker/dependencychecker.py
+++ b/z3c/dependencychecker/dependencychecker.py
@@ -501,7 +501,6 @@ def includes_from_django_settings(path):
                 "Found possible packages in Django-like settings file %s:",
                 settingsfile)
             logger.debug(found)
-            # import pdb;pdb.set_trace()
             if 'test' in settingsfile:
                 # testsettings.py, for instance.
                 test_modules += found
--- a/z3c/dependencychecker/dependencychecker.py
+++ b/z3c/dependencychecker/dependencychecker.py
@@ -637,29 +637,36 @@ def main():
 
     print_unused_imports(unused_imports)
 
+    all_install_imports = (
+        install_imports +
+        zcml_imports +
+        generic_setup_required +
+        django_settings_imports +
+        fti_imports)
     install_missing = filter_missing(
-        install_imports + zcml_imports + generic_setup_required +
-        django_settings_imports + fti_imports,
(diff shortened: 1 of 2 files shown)


z3c/dependencychecker/dependencychecker.py
```diff
@@ -501,9 +514,9 @@ def imports_from_doctests(path):
     for path, dirs, files in os.walk(path):
         for filename in [os.path.abspath(os.path.join(path, filename))
                          for filename in files
-                         if fnmatch.fnmatch(filename, '*.txt')
-                         or fnmatch.fnmatch(filename, '*.rst')
-                         or fnmatch.fnmatch(filename, '*.py')]:
+                         if fnmatch.fnmatch(filename, '*.txt') or
+                         fnmatch.fnmatch(filename, '*.rst') or
```
reinout: Is this pep8-suggested? I like the original a bit more, visually, but I know that "nice visual effect" isn't always the best reason :-)

reinout:
Looks OK. A bit of cleanup is nice :-)

I've added one question in the PR.

z3c/dependencychecker/dependencychecker.py
mauritsvanrees: Yes.  the pep8 checker says: `W503 line break before binary operator`.
Ah, that is a recent change that is added to the pep8 guidelines (but not visible on the official pep yet).
See the links in this issue: https://github.com/PyCQA/pycodestyle/issues/498

reinout: So.... "I was right all along"? :-)

mauritsvanrees: Well... you were right all along, but not anymore. :-) My change follows the new rules.

reinout: You're sure? If I read https://hg.python.org/peps/rev/3857909d7956, it seems like you should break before the binary operator. If the break is before the "or/and", then the "or/and" is the first one on the line, right?

mauritsvanrees: Argh, I am misreading it. I have reverted (removed) this commit.

BTW, the pep8 has been renamed to pycodestyle.

reinout:
Maurits: I added you as collaborator, btw.

PR #44, opened 2017-06-21
Django improvements

reinout:
I'm using z3c.dependencychecker now on a fairly large django codebase. After a few tweaks, it works quite well!

Here are the tweaks :-)


reinout:
Merci

PR #45, opened 2017-08-05
Replace bootstrap.py with requirements.txt

gforcada:
It is more reliable (pip) to install setuptools and zc.buildout with problems.


reinout:
I'll merge it.

Question: is this the new standard way as used by plone and so?

Question 2: you probably do it inside a virtualenv. Does virtualenv's `bin/` dir conflict with buildout's?

gforcada:
@reinout thanks for merging!

Regarding questions:

- we are moving on that direction, see https://github.com/plone/Products.CMFPlone/issues/2001
- they shouldn't as long as you only install setuptools and zc.buildout with pip and the rest with buildout and the versions are kept in sync

Regarding the second point, that should be the only usage of pip on a typical plone installation

Issue #32, opened 2015-08-28
Make sure it runs on python 3


reinout:
TODO: print statements adjustments and so.

gforcada:
There's a bigger elephant on the room: ``importchecker.py`` is using the ``compiler`` module that got removed on python 3, I guess that maybe we need to move that code to use the ``ast`` one?

reinout:
Ouch, looks that way, yes. I personally haven't used ast (or the compiler module) a lot. Most of the importchecker.py code is directly copied from a zope helper function somewhere.

gforcada:
Should we keep making releases on the 1.x series with a command line switch to use either of the implementations and remove the old one for a 2.0 release with python 3 support? 🤔

reinout:
Just remove the old implementation. If someone needs the old version, they can just use an old release.

Just making sure: the new implementation is for both python 2.7 and 3.x?

gforcada:
@reinout ok, I will remove them, I wanted to keep a command line switch for a while so we can easily compare the output of the two implementations out in the wild (i.e. get all 300 Plone packages, run dependencychecker on all of them with the two implementations, store that ouput and compare the two versions).

I'm sure that there will be plenty of corner cases.

And yes, the new implementation is working fine for 2.7 and last time I checked there was only 1 single test that was failing on 3.6, a really minor one though.

reinout:
Well, I assumed it would be quickest/easiest to just remove the old implementation.
If it is easy to keep it around for debug purposes, that's fine with me, too.

reinout:
2.0 runs on python 3!

PR #46, opened 2017-08-05
Add coverage integration

gforcada:
@reinout if you agree on these changes, first of all https://coveralls.io should be enabled, to do so one needs to log in on coveralls.io and click around.

My idea is to try to tackle the rewrite of importchecker.py at some point to use the ast module so it can be more reliable. Actually, now that I realize, this is also a blocker for z3c.dependencychecker itself to be usable in Python 3 (see #32)

The main idea behind this is that plone, on the way to become a python 3 citizen it should also clean up its completely tangled dependency graph. If we don't get all the dependencies on setup.py first, that's nearly to impossible, so we need z3c.dependencychecker to help us on that!

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,7 +4,6 @@ branch = True
 [report]
 omit =
     */migrations/*
-    */test*
     /Library/*
     *buildout/eggs/*
     /usr/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,6 @@
+[run]
+branch = True
+
 [report]
 omit =
     */migrations/*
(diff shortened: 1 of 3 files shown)


reinout:
Looks good!

Using it to tidy up plone: that's exactly what it is made for. One of the first usecases was tidying up some zope/grok libraries. Somewhere in 2009 :-) I've recently started using it for a big django project.

Using the 'ast' module: great idea.

Note that I don't mind at all if the functionality to detect unused imports gets deleted. We have pyflakes for that, right? Otoh, if it is easy to keep around: fine.

gforcada:
@reinout thanks for merging the pull requests, I'm starting slowly, I'm on parental leave, to rewrite that script, as soon as I have something I will create a WIP pull request to quickstart the discussion.

PR #49, opened 2017-08-27
Rewrite tests

### LINKED ISSUE #47: rewrite tests?
The more I'm looking at the code, or rather how to refactor and rewrite it from scratch (picking up what's still interesting), the more I want to look at the tests...

Which leads me to a situation where I would actually like to rewrite them to be regular unittest tests, which would mean that we could use something like pytest or the likes to run, rather than ancient and barely developed z3c.testsetup and zope.testing.

@reinout @mauritsvanrees What do you think?

gforcada:
Fixes #47


reinout:
The rewrite to pytest looks fine.

What I'm not yet happy with is the pip/buildout difference. There are now two. And instead of "bin/test" I've got a long pytest command line. (I'm myself debating which combination of pip/buildout to use for future projects, so that's why I'm asking).

gforcada:
@reinout the long pytest command line can basically be only "pytest z3c" and that's enough. The extra bits are only needed to get the coverage report.

As for pip/buildout, we only keep 3 things in buildout right now: omelette (probably useful?),  console_scripts (no longer needed?) and code analysis (which sadly has a hard dependency on buildout).

I would like to move the last one (plone.recipe.codeanalysis) to be a plain command line script, just like zest.releaser or flake8, but that can take a while...

Regarding this pull request, is the code style ok?

gforcada:
see #50

reinout:
Code style is fine.
Personally, I'd replace the sysargv-faker with the `mock` library, but that's because I use it the whole time. For one thing only it isn't needed :-)
I'll merge #50 and do some more thinking on pip/buildout.

Perhaps we have add a note in the readme regarding pip/buildout?

PR #54, opened 2017-08-31
WIP: rewrite

gforcada:
@reinout @mauritsvanrees would you mind giving a first review of the last commit on this branch?

That's my attempt to use ``pkg_resources`` to get a random package dependencies. There are no tests whatsoever, as it's stated, is just a proof of concept, that so far on the tests I've done is working fine.


z3c/dependencychecker/dependencychecker.py
```diff
@@ -655,6 +655,9 @@ def set_log_level(verbose):
 
 
 def main():
```
reinout: You *could* just move this to `main.py`?

z3c/dependencychecker/main.py
```diff
@@ -0,0 +1,79 @@
...
+    logger.info('path used: %s', path)
+    return path
+
+
+def get_package_dependencies(path):
+    if not has_setup_py_file(path):
+        logger.info('No setup.py file found! on {0}'.format(path))
+        sys.exit(1)
+
+    new_path, filename = get_egg_info_folder(path)
```
reinout: Egg info: do we need to support (or switch to) other approaches? wheels have another dir/file if I'm correct.

z3c/dependencychecker/main.py
```diff
@@ -0,0 +1,79 @@
...
+    if not new_path:
+        logger.info('No egg-info folder found! on {0}'.format(new_path))
```
reinout: http://reinout.vanrees.org/weblog/2015/06/05/logging-formatting.html

So the best way is to stick to ye olde `%s` :-)

z3c/dependencychecker/main.py
```diff
@@ -0,0 +1,79 @@
...
+        sys.exit(1)
+    path = new_path
+    package_name = filename[:-len('.egg-info')]
+
+    working_set = pkg_resources.WorkingSet()
+    working_set.add_entry(path)
+    new_requirement = pkg_resources.Requirement.parse(package_name)
```
reinout: I notice I find it a bit hard to understand what "new" is. `new_requirement`, `new_path`.
Names like `our_package_name`, `our_own_requirement`, `our_egg_info_path` (just to give an idea) make it clearer to follow what happens.

z3c/dependencychecker/main.py
```diff
@@ -0,0 +1,79 @@
...
+        only_extra = set(all_with_extra) - set(dependencies)
+        extras_list[extra] = only_extra
+
+    from pprint import pprint
+    pprint(dependencies)
+    pprint(extras_list)
+
+
+def has_setup_py_file(path):
+    return bool([x for x in os.listdir(path) if x == 'setup.py'])
```
reinout: Much shorter: `return 'setup.py' in os.listdir(path)`

z3c/dependencychecker/main.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def get_egg_info_folder(path):
+    def get_egg_folder(test_path):
+        return [
```
reinout: You return a list, not a specific folder. Is that right?

reinout: And: isn't a `glob.glob('*.egg-info')` handier/quicker?

z3c/dependencychecker/main.py
```diff
@@ -0,0 +1,79 @@
...
+            x
+            for x in os.listdir(test_path)
+            if x.endswith('.egg-info')
+        ]
+
+    egg_info = get_egg_folder(path)
```
reinout: `egg_info` doesn't sound like a filename/path, it sounds more like some pkg_resources object

z3c/dependencychecker/main.py
```diff
@@ -0,0 +1,79 @@
...
+    if not egg_info:
+        logger.info('Try to append src/ to the current path %s', path)
+        path = os.path.join(path, 'src')
+        egg_info = get_egg_folder(path)
+
+    if egg_info:
+        return path, egg_info[0]
```
reinout: You use it as `new_path, filename` earlier up in the file. It is more readable if you pick one/two good explicit name(s) and use the same in both places.

z3c/dependencychecker/main.py
```diff
@@ -0,0 +1,79 @@
...
+    return False
```
reinout: `new_path, filename = get_egg_info_folder(path)` will raise an error if you feed it a one-element `False`

gforcada:
Thanks for the feedback! I will go over it during the next days.

z3c/dependencychecker/dependencychecker.py
gforcada: Yeah, that's my idea (and plenty more actually) I just wanted to have a minimal version to show to you first.

z3c/dependencychecker/main.py
gforcada: So far we only allow to analyze source distributions that have been "enabled", i.e. ``python setup.py develop`` be it with buildout, pip or whatever.

Do we want to expand that to more approaches?

I was already planning to reformat the code, so we can have a generic entry point that looks at which kind of distribution are we getting and then call the right parser/extractor.

z3c/dependencychecker/main.py
gforcada: Yup, I will fix that up.

z3c/dependencychecker/main.py
gforcada: Yeah this last chunk of code was copy & pasted directly from an interactive terminal 🙂 I will change them to proper names, function names as well and some docstrings.

z3c/dependencychecker/main.py
gforcada: I knew it had to be easier, I was doing some list comprehensions some lines above/below and I lazily copy/pasted it 😅

z3c/dependencychecker/main.py
gforcada: I barely used ``glob`` thanks for pointing it out!

gforcada: And yes, to reply backwards, it returns a list, although only with one item hopefully, I will refactor that to be more specific.

z3c/dependencychecker/main.py
gforcada: I will review the naming, as I said, it was a quick and dirty put all together so it can be discussed.

z3c/dependencychecker/main.py
gforcada: Yup, good idea 👍

z3c/dependencychecker/main.py
gforcada: Spot on! I will fix that as well

gforcada:
@reinout @mauritsvanrees that's quite a lot of code already, but just so you can see the direction in which I'm heading to.

reinout:
@gforcada, I think this PR can be closed now, as everything is in 2.0?

gforcada:
@reinout we can close it yes, but there is more stuff that is on it that is not on 2.0. I will send some pull requests later.

PR #56, opened 2017-09-26
Small cleanups

gforcada:
While working on #54 I kept some small cleanups that are just noise for the rewrite, so let's better merge them before the big one.

As for #54 status, as I don't have much time nor focus is a bit stalled, but improving it every now and then :-) I hope to push something useful before the end of October, as then I will start working again and thus my spare time will be even more scarce :-/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 sudo: false
+cache: pip
 python:
   - 2.7
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ cache: pip
 python:
   - 2.7
 install:
+  - rm -rf /home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/setuptools-36.0.1.dist-info
   - pip install -r requirements-dev.txt
 # force failure of code analysis
   - buildout code-analysis:return-status-codes=True
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -20,8 +20,6 @@ pycparser==2.18
 pyOpenSSL==17.2.0
 pytest==3.2.1
 pytest-cov==2.5.1
-pytest-cover==3.0.0
-pytest-coverage==0.0
 requests==2.18.4
 six==1.10.0
 urllib3==1.22
--- a/z3c/dependencychecker/tests/test_integration.py
+++ b/z3c/dependencychecker/tests/test_integration.py
@@ -60,4 +60,4 @@ def test_highlevel_integration(capsys, fake_project):
         with mock.patch.object(sys, 'argv', arguments):
             main()
             out, err = capsys.readouterr()
-            assert out == MAIN_OUTPUT
+            assert MAIN_OUTPUT in out


.travis.yml
```diff
@@ -3,6 +3,8 @@ sudo: false
 python:
   - 2.7
 install:
+  - virtualenv .
```
reinout: I thought travis gives you a virtualenv by default? I mean, I remember deactivating the virtualenv in one of my `.travis.yml` files in order to access some system packages :-)

README.rst
```diff
@@ -50,15 +50,6 @@ It checks the following locations:
 - Python files, ``.txt`` and ``.rst`` files for imports in doctests.
 
 
-Note on running the tests
```
reinout: Isn't this paragraph still valid for local development?

z3c/dependencychecker/__init__.py
```diff
@@ -1 +1,4 @@
 # -*- coding: utf-8 -*-
+import logging
+
+logger = logging.getLogger('dependencychecker')
```
reinout: I dislike having stuff in an `__init__` unless absolutely necessary.

z3c/dependencychecker/dependencychecker.py
```diff
@@ -25,10 +25,9 @@
 import pkg_resources
 
 from z3c.dependencychecker import importchecker
+from z3c.dependencychecker import logger
 
 
-logger = logging.getLogger(__name__)
```
reinout: Why this change? It is a very common mechanism.

.travis.yml
gforcada: If you try this branch on travis without the last commit it fails with a setuptools version conflict, see: https://travis-ci.org/reinout/z3c.dependencychecker/builds/279908297

gforcada: Looking at it in more detail it seems that:

- travis has setuptools **36.3.0** installed
- it installs (via requirements.txt) version **33.1.1**
- pip detects 36.3.0 installed and uninstalls it (or so it says...)
- on the final overview of the pip install -r requirements-dev.txt it says that setuptools **36.0.1** is installed
- then buildout starts and has a conflict between 33.1.1 and 36.0.1

No idea why this 36.0.1 comes from...

README.rst
gforcada: Well, tests are no longer that sensitive, at least that's my aim 🙂 

With the current travis setup, that one should replicate for local development, the environment is already isolated.

z3c/dependencychecker/__init__.py
gforcada: @reinout fair point, I put it there as a way to avoid having to create a 2 lines module for it, another solution that I don't like is adding it to main.py or any other random module, as then you keep importing that module from everywhere just for the logger.

I'm fine with creating a logger.py if you think is best.

z3c/dependencychecker/dependencychecker.py
gforcada: Yes, and I use (and test) it quite a lot! What I did not like is that to get the logger I had to import a module named ``dependencychecker``. When I think about getting the package's logger I either think about that is on ``__init__.py`` or on some utility module, not on a random module that happens to have it.

Does that make sense?

reinout: You don't need to import some logger, just put the `logger = logging.getLogger(__name__)` line in every file (and a `import logging` import, obviously).

The `.getLogger()` statement doesn't result in an instance you're only allowed to have one from.

README.rst
reinout: The you should point at the `.travis.yml` file and say that you should replicate that for local development.

I think it is easier to just keep the current text (until the tests are actually less sensitive).

z3c/dependencychecker/__init__.py
reinout: You don't need a `logger.py`! You can just have a `logger = logging.getLogger(__name__)` line in every file. That's the default in all my projects (and also in django, I believe).

z3c/dependencychecker/dependencychecker.py
gforcada: Yeah, I was doing it just so that I don't have to keep creating it on every file, if you prefer this way, that's fine with me.

README.rst
gforcada: ok

gforcada:
@reinout now that's *interesting*... if I manually install setuptools 33.1.1 it works, but not when it comes from requirements.txt...

We could bump setuptools version as well as a solution. Having to keep the version in 3 different files is already too much (requirements.txt, buildout.cfg and now .travis.yml as well).

I was thinking about removing it from requirements.txt so then we are still at 2 places only, but then we will not have repeatable local environments, as we will not pin which setuptools version you need...

gforcada:
Diving a bit more into it, it seems that https://github.com/travis-ci/travis-ci/issues/8493 could be what's happening to us as well.

reinout:
@gforcada: that travis issue sure looks like the problem. Leaving your "virtualenv enable" in there for now is a fine solution. We can easily remove it when it is no longer needed.

reinout:
Or perhaps a brute-force `rm -rf that-weird-egg-info-file`?

Does it help maintenance to unpin setuptools everywhere? Instead of pinning a specific version? That way you don't need to change the version in multiple places.

gforcada:
@reinout indeed, removing that stray folder fixes the downgrade problem.

Regarding unpinning setuptools... I would rather keep it pinned as I really dislike encountering problems all of a sudden due to some version being unexpectedly bumped and causing havoc.

My motto is usually: *keep things repeatable, so you have a solid/known base to work on*

gforcada:
@reinout I kept the message on README, so I guess this could go in if you wish. 🙂

PR #60, opened 2017-12-12
DottedName class

gforcada:
This is the first meaningful on a series of pull requests (as I like to keep the git history clean I will make one at a time if that's ok @reinout @mauritsvanrees ).

This pull requests adds a new module with only a single class: **DottedName**.

This is mostly a wrapper around either an import statement or a requirement (from setup.py).

Together with the PackageMetadata (see #54 for the whole diff) are the two most important building blocks, although this one does not have that much code actually.

Question: should I add some docstrings here and there or is the code explanatory enough?

Note that this code is only being tested, current code is not using it in any way, all but the last pull requests will use the same approach, so that we can keep merging them but not affecting a quick fix release or any development that could happen.

### DIFF
--- a/z3c/dependencychecker/dotted_name.py
+++ b/z3c/dependencychecker/dotted_name.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
+from functools import total_ordering
 
 
+@total_ordering
 class DottedName(object):
 
     def __init__(
@@ -40,3 +42,14 @@ class DottedName(object):
     @staticmethod
     def _is_namespaced(namespaces):
         return bool(len(namespaces) - 1)
+
+    def __lt__(self, other):
+        if not isinstance(other, DottedName):
+            return NotImplemented
+        return self.name < other.name
+
+    def __eq__(self, other):
+        if not isinstance(other, DottedName):
+            return NotImplemented
+
+        return self.safe_name == other.safe_name
--- a/z3c/dependencychecker/dotted_name.py
+++ b/z3c/dependencychecker/dotted_name.py
@@ -11,6 +11,9 @@ class DottedName(object):
         is_test=False,
     ):
         self.name = name
+        self.safe_name = self._get_safe_name(name)
+        self.namespaces = self._get_namespaces(self.safe_name)
+        self.is_namespaced = self._is_namespaced(self.namespaces)
         self.file_path = file_path
         self.is_test = is_test
 
@@ -23,3 +26,17 @@ class DottedName(object):
             requirement.project_name,
             file_path=file_path,
         )
+
+    @staticmethod
+    def _get_safe_name(name):
+        safe_name = name.lower().replace('-', '_')
+        return safe_name
+
+    @staticmethod
+    def _get_namespaces(safe_name):
+        parts = safe_name.split('.')
+        return parts
+
+    @staticmethod
+    def _is_namespaced(namespaces):
+        return bool(len(namespaces) - 1)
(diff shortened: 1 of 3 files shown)


z3c/dependencychecker/dotted_name.py
```diff
@@ -0,0 +1,77 @@
...
+        return self.safe_name == other.safe_name
+
+    def __repr__(self):
+        return '<DottedName {0}>'.format(self.name)
+
+    def __hash__(self):
+        digest = hashlib.sha256(self.safe_name).hexdigest()
+        return int(digest, 16) % 10**8
+
+    def __contains__(self, item):
```
reinout: This is the only place where I'd like a short docstring. What does containment mean here (or rather: how do you want to use it).

Looks like you're making something that is nicely pythonic, btw! Equality, hash, etc. Good job!

z3c/dependencychecker/dotted_name.py
```diff
@@ -0,0 +1,77 @@
...
+    def __lt__(self, other):
+        if not isinstance(other, DottedName):
+            return NotImplemented
+        return self.name < other.name
+
+    def __eq__(self, other):
+        if not isinstance(other, DottedName):
+            return NotImplemented
+
+        return self.safe_name == other.safe_name
```
mauritsvanrees: For equality you compare the safe_name, but for 'lighter than' you compare the name. That seems wrong: a dottedname of `Plone` would then be equal to `plone` (lowercase they are the same), but at the same time `Plone` would be lighter than `plone`.
I don't know if that is a problem in practice.
I would suggest changing it, unless you have a good reason not to (maybe the code was already like this).

gforcada: Given the test package used for testing this is how I had to put them to make it work.

If memory serves me well, the equality is to handle pypi mixed case and imports being different case and the sorting is due to alphabetically sorting the output:

https://github.com/reinout/z3c.dependencychecker/blob/master/z3c/dependencychecker/tests/test_integration.py#L16-L27

If we used safe_name then Products.GenericSetup would not be at the top but rather on the middle. I actually made the change and then pytest told me to undo it 🙂

z3c/dependencychecker/dotted_name.py
gforcada: There is a database class that actually does the heavy lifting, and there I use sets and list comprehensions to filter, so having all these dunder methods really keeps code simpler.

I will write a docstring on that one.

z3c/dependencychecker/dotted_name.py
gforcada: @mauritsvanrees although I merged I still want to hear back from you if you are ok with my explanation

mauritsvanrees: I would not mind having `Products.GenericSetup` end up in the middle. That makes it fit the case insensitive alphabetic sorting that we prefer in Plone.
I just worry that somewhere down the line it might give problems if `a==b ` and `a<b` are both true.
If it works for us, it may be no problem.

gforcada: I will keep that in mind, but so far it works just like the current implementation.

Ao although we could change that now, I would like to:
- have all #54 merged (in the small pull requests that I'm sending)
- create a 2.0.dev0 release and try it in the wild
- keep ironing out corner cases that have been missed
- make a final 2.0 release 🙂

reinout: Fine with me.

PR #61, opened 2017-12-12
Package Metadata

gforcada:
Second one, and a crucial one, extracting all the needed info from the .egg-info dir, while trying to stay as much away from it as possible (i.e. using APIs as much as possible rather than direct file access).


z3c/dependencychecker/package.py
```diff
@@ -0,0 +1,178 @@
...
+                    '.egg-info folder found at %s folder',
+                    path,
+                )
+                return path
+
+        logger.error('.egg-info folder could not be found')
+        sys.exit(1)
+
+    @staticmethod
+    def _check_egg_in_folder(path):
```
reinout: I'd call it `_find_egg_info_in_folder()`. You're not checking anything, you're searching/finding. And "egg_info" instead of "egg" is a tad more explicit.

z3c/dependencychecker/package.py
```diff
@@ -0,0 +1,178 @@
...
+        """Use pkg_resources API to enable the package being analyzed
+
+        This, enabling the package in pkg_resources, allows to extract
+        information from .egg-info folders, for example, its requirements.
+        """
+        working_set = pkg_resources.WorkingSet()
+        working_set.add_entry(self.package_dir)
+        return working_set
+
+    def _get_package_from_working_set(self):
```
reinout: "package" is probably the original terminology? I'm asking because you return a "distribution". And the docstring mentions a "requirement". Three terms for the same thing.

I know from buildout that it can be quite confusing whether you have to deal with requirements or installed packages or.... => perhaps take a quick look if one or two of the names need changing.

z3c/dependencychecker/package.py
gforcada: 👍 thanks, naming *is hard* as we all know. I tried to be as consistent as possible, but there's always room for improvement

z3c/dependencychecker/package.py
gforcada: Yup, that was way too much, I removed distribution and changed it to package, I keep Requirement on the docstring as that's the object class one gets from pkg_resources.

And while at it I decided to rename two methods:
``_get_package_from_working_set`` -> ``_get_ourselves_from_working_set``
``_generate_working_set_with_package`` -> ``_generate_working_set_with_ourselves``

Looking at the method names it seemed to imply that an argument of them was a random package that you should give, but it is rather the very same package that we are dealing with, so I changed ``package`` to ``ourselves``

reinout: Much clearer!

PR #62, opened 2017-12-13
Extractors

gforcada:
That's probably where more corner cases will be found, I created a class for each source of imports.

I tried to avoid regexes and rather use ast and lxml to parse files.

Let's see what's your opinion 🙂

### DIFF
--- a/z3c/dependencychecker/modules.py
+++ b/z3c/dependencychecker/modules.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+import os
+import re
+
+
+TEST_REGEX = r"""
+{0}    # Needs to start with the path separator
+\w*    # There could be some characters (or none)
+test   # has to contain test on it
+s?     # ensure plurals are handled as well
+(      # Needs to end with either a path or a file extension:
+{0}    # path separator
+|      # OR
+\w*    # some possible characters
+\.     # a dot
+\w+    # extension
+)
+""".format(os.sep)
+
+TEST_IN_PATH_REGEX = re.compile(TEST_REGEX, re.VERBOSE)
+
+
+class BaseModule(object):
+
+    def __init__(self, package_path, full_path):
+        self.path = full_path
+        self._relative_path = self._get_relative_path(package_path, full_path)
+        self.testing = self._is_test_module()
+
+    @staticmethod
+    def _get_relative_path(package_path, full_path):
+        return full_path[len(package_path):]
+
+    def _is_test_module(self):
+        return bool(re.search(TEST_IN_PATH_REGEX, self._relative_path))
+
+    @classmethod
+    def create_from_files(cls, top_dir):
+        raise NotImplementedError
+
+    def scan(self):
+        raise NotImplementedError
--- a/z3c/dependencychecker/modules.py
+++ b/z3c/dependencychecker/modules.py
@@ -40,3 +40,41 @@ class BaseModule(object):
 
     def scan(self):
         raise NotImplementedError
+
+
+class PythonModule(BaseModule):
+
+    @classmethod
+    def create_from_files(cls, top_dir):
+        """Find all python files in the package
+
+        For that it gets the path to where top_level.txt points to,
+        which is not always a folder:
+        - folder example: z3c.dependencychecker itself
+        - file example: flake8-isort or any other single file distribution
+
+        Return this very same class, which would allow to call the scan()
+        method to get an iterator over all this file's imports.
+        """
+        if top_dir.endswith('.py'):
+            yield cls(top_dir, top_dir)
+            return
+
+        for path, folders, filenames in os.walk(top_dir):
+            if '__init__.py' not in filenames:
+                # Don't descend further into the tree.
+                # Clear folders variable in-place.
+                folders[:] = []
+                continue
+
+            for filename in filenames:
+                if filename.endswith('.py'):
+                    yield cls(
+                        top_dir,
+                        os.path.join(path, filename),
+                    )
+
+
+MODULES = (
+    PythonModule,
+)
--- a/z3c/dependencychecker/modules.py
+++ b/z3c/dependencychecker/modules.py
@@ -276,9 +276,54 @@ class GSMetadata(BaseModule):
                 )
 
 
+class PythonDocstrings(PythonModule):
+    """Extract imports from docstrings found inside python modules
+
+    There are some projects that rather than having separate files to keep the
+    tests, they add the inline with the code as docstrings,
+    either at class or method/function level.
+    """
+
+    NODES_WITH_DOCSTRINGS = (
+        ast.Module,
+        ast.ClassDef,
+        ast.FunctionDef,
+    )
+
+    def scan(self):
+        for node in ast.walk(self._get_tree()):
+            if isinstance(node, self.NODES_WITH_DOCSTRINGS):
+                docstring = ast.get_docstring(node)
+                for dotted_name in self._parse_docstring(docstring):
+                    yield dotted_name
+
+    def _parse_docstring(self, docstring):
+        if not docstring:
+            return
+
+        for line in docstring.split('\n'):
+            code = self._extract_code(line)
+            if code:
+                try:
+                    tree = ast.parse(code)
+                except SyntaxError:
+                    return
+
+                for node in ast.walk(tree):
+                    for dotted_name in self._process_ast_node(node):
+                        yield dotted_name
+
+    @staticmethod
+    def _extract_code(line):
+        if '>>>' in line:
+            position = line.find('>>>') + 3
+            return line[position:].strip()
+
+
 MODULES = (
     PythonModule,
     ZCMLFile,
     FTIFile,
     GSMetadata,
+    PythonDocstrings,
 )
--- a/z3c/dependencychecker/modules.py
+++ b/z3c/dependencychecker/modules.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from xml.etree import ElementTree
 from z3c.dependencychecker.dotted_name import DottedName
 import ast
 import os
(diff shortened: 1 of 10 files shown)


gforcada:
*side note:* the more I look at github commits tab the weirder it gets... I keep rebasing my commits and refactoring them so if you look at the terminal you will see them properly sorted, but github uses some other sort of algorithm to sort them and now you have a completely mixed up history there 😕

reinout:
I'm very, very basic in my use of git features. "Rebase" and so always sounds like "dark magic" that can get you arrested by the Spanish Inquisition...

gforcada:
Spanish inquisition 😆 ... being Catalan, yes, it is possible yes 🙂 

I'm quite a git nerd myself, and mastering rebase is a super important tool for me, specially once you know that ``git reflog`` exists, so when you screw up, you can still get back to where you were 👍

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def scan(self):
+        raise NotImplementedError
+
+
+class PythonModule(BaseModule):
+
+    @classmethod
+    def search_files(cls, top_dir):
+        """Finds all python files on the package
```
reinout: "on" => "in"? "inside"? "of"?

reinout: And write it as something like this: "Return top_dir/abs_file_path tuple for every python file in the package"

- Write it as an order ("find!", "return!") instead of as a description ("returns", "finds").

- "Finds" doesn't day what actually happens when it finds something.

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+        raise NotImplementedError
+
+    def scan(self):
+        raise NotImplementedError
+
+
+class PythonModule(BaseModule):
+
+    @classmethod
+    def search_files(cls, top_dir):
```
reinout: `def instantiate_from_package_dir()` ?

z3c/dependencychecker/modules.py
reinout: Actually, it returns class instances

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+
+    @staticmethod
+    def _get_relative_path(package_path, full_path):
+        return full_path[len(package_path):]
+
+    def _is_test_module(self):
+        return bool(re.search(TEST_IN_PATH_RE, self._relative_path))
+
+    @classmethod
+    def search_files(cls, top_dir):
```
reinout: See comment on `PythonModule` below regarding name.

reinout: You'll have to make clearer what the flow is. What does this method do? What does `scan()` do? Which gets called first? There are quite some methods in the class.

I haven't looked at the rest of the code yet, but my *current* guess (and thus also my current naming suggestion) would be:

- `search_files()` is the objects-creating method. Iterate through something and return instantiated `BaseModel` subclasses. An `instantiate_from_dir/package/file/whatever()` might be clearer. Or `create_....()`?

- `scan()` is what then gets called (by what? I would expect it to be called in the `__init__()`?) to read the file and extract info from it. Would `.extract_info()` be a clearer name? Depends also on what it actually does.

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+                continue
+
+            for filename in filenames:
+                if filename.endswith('.py'):
+                    yield cls(
+                        top_dir,
+                        os.path.join(path, filename),
+                    )
+
+    def scan(self):
```
reinout: What does this do? Where does it get called?

Sounds like `get_mentioned_dotted_names()` or so?

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+        'adapter': ('for', 'factory', 'provides', ),
+        'utility': ('provides', 'component', ),
+        'browser:page': ('class', 'for', 'layer', ),
+        'subscriber': ('handler', 'for', ),
+        'securityPolicy': ('component', ),
+        'genericsetup:registerProfile': ('provides', ),
+    }
+
+    @classmethod
+    def search_files(cls, top_dir):
```
reinout: Yeah, something like `instantiate_from_package()` sounds nicer and nicer :-)

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+                elif name in ('klass', 'schema'):
+                    if node.text:
+                        yield DottedName(
+                            node.text.strip(),
+                            file_path=self.path,
+                            is_test=self.testing,
+                        )
+
+
+class GSMetadata(BaseModule):
```
reinout: Add a docstring (or comment) that this is about plone's generic setup stuff.
Someone from the django world will be puzzled, otherwise, when reading this code :-)

reinout:
Comment not directly related to this PR: I really like reviewing your pull requests. You use a different approach than that I'm used to and I'm learning from that :-)

And... your code is very neat. Pep8, pyflakes and so.

reinout:
Comment that *is* related to this pull request: you don't use logging.
I've noticed that I very much needed logging when working on z3c.dependencychecker. "Why isn't it picking up that I'm using package x.y.z?" and "why does it think I depend on x.y.z?".

So: I'm not opposed to a bunch of `logger.debug()` statements.

gforcada:
@reinout thanks for the compliment 😊 as I have to review a lot of code at work, I tend to get quite picky on things, so ``bin/code-analysis && pytest z3c`` are my precious tools to move forward and backwards through commits to ensure everything is fine

regarding logging, I did add them but later on. As you see here I'm using generators, so I placed the logging on the consumers,  this way with just a single line of logging we log all these classes, would that be enough?

gforcada:
@reinout one more thing regarding logging, if you checkout #54 branch and run ``dependencychecker -v`` you will see the logging yourself

reinout:
Ah, so the logging is only missing because you've split up #54 into separate pull requests and will be added in a later PR? Ok, clear.  And logging on the consumers is fine.

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+import os
+import re
+
+
+TEST_REGEX = r"""
+{0}    # Needs to start with the path separator
+\w*    # There could be some characters (or none)
+test   # has to contain test on it
+s?     # ensure plurals are handled as well
+(      # Needs to end with either a path or a file extension:
```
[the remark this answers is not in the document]
gforcada: Good point, so far on all projects I've ever looked, there has been always a ``test``/``tests`` folder.

On [pytest docs](https://docs.pytest.org/en/latest/goodpractices.html), they mention two layout techniques (test in or out of tree) and for both they name the folder containing the tests as ``tests``.

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+{0}    # path separator
+|      # OR
+\w*    # some possible characters
+\.     # a dot
+\w+    # extension
+)
+""".format(os.sep)
+
+TEST_IN_PATH_RE = re.compile(
```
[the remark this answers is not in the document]
gforcada: Good points, I amended both.

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+    TEST_REGEX,
+    re.VERBOSE,
+)
+
+
+class BaseModule(object):
+
+    def __init__(self, package_path, full_path):
```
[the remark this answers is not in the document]
gforcada: I see, sorry for the confusion. Let me tell its story :bonfire:

Up until a week or so there was only a single argument, ``path`` stored in ``self.path``.

But then while working on the regex I noticed that as I was passing the full path (i.e. self.path) to the file, if the git checkout of the package being analyzed happened to be on ``/home/gil/source/tests/my.package``, then every single module analyzed would be marked as a test module.

For that I introduced a second argument on the init method (package_path) so that I could get only the relevant part of the filesystem path and thus avoid those problems.

If you look at the usage of ``_relative_path`` is only being used in ``_is_test_module`` method, nowhere else.

That's why I decided to keep it as a private variable and keep the other variable as ``path``.

Does that make more sense now? Probably a docstring on ``__init__`` explaining that would help?

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+        self.path = full_path
+        self._relative_path = self._get_relative_path(package_path, full_path)
```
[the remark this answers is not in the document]
gforcada: It is used, only once though, on ``_is_test_module`` :-)

z3c/dependencychecker/modules.py
```diff
@@ -0,0 +1,398 @@
...
+        self.testing = self._is_test_module()
+
+    @staticmethod
+    def _get_relative_path(package_path, full_path):
```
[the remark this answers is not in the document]
gforcada: At some point it started bothering me as well, but as you say, I kept it for consistency.

I'm completely open to any suggestions though, I will have a look at cached-property. I was thinking of using plain ``@property`` but I was not so sure if my usage is what they are meant for. I have barely used them...

z3c/dependencychecker/modules.py
gforcada: I'm starting to regret the way I have split the branch... :-/

This commit should shed some light: https://github.com/reinout/z3c.dependencychecker/pull/54/commits/eba1ceb40517e52717876b8f7d641b85e2fcd2c1

In short:

``` 
for module_obj in MODULES:
    for source_file in module_obj.search_files(top_folder):
        self.imports.add_imports(source_file.scan())
```
At the very bottom of modules.py there is a tuple with all the classes defined there. This is looped through (see above) and on each of them the ``search_files`` method is called which returns an iterator with all the files for that given specific extractor that can be found.

For each file found, a new object from the same class is returned and when called on the ``scan`` method it returns an iterator giving you all imports (if any) that have been found on that file.

I thought that a ``@classmethod`` rather than some external method/function would be nicer as it allows us to wrap both functionality (searching specific files and searching imports on those specific files) in a single class. Probably it would make sense to describe this *protocol* on the module global docstring, so that another contributor willing to support whatever files (markdown?) would know what to do.

As for the method names, I'm completely open to change them, but I guess this explanation was needed first.

z3c/dependencychecker/modules.py
gforcada: See above my large answer, although still open to change them if you still think the name change is needed.

z3c/dependencychecker/modules.py
gforcada: Updated all of them and added a docstring on all classes.

z3c/dependencychecker/modules.py
gforcada: See the large answer some comments above

z3c/dependencychecker/modules.py
gforcada: Done

gforcada:
@reinout let me know what you think about my explanations on your (great btw!) review

z3c/dependencychecker/modules.py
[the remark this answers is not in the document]
reinout: Leave it for now, we can always refactor it later. Otherwise we're probably making it unnecessary hard to get the total PR merged :-)

z3c/dependencychecker/modules.py
reinout: "search" doesn't sound so very active to me. I'd like something with "create" or "instantiate", that makes it clearer that this is the main object-instantiating method.

What about `create_from_files()`? Then `scan()` can stay. You create objects from files and then ask them to scan themselves. Sounds clear enough.

gforcada:
@reinout let me know what you think about my explanations on your (great btw!) review

PR #63, opened 2017-12-14
Database and Package

gforcada:
Next one then 🙂 

In this one two classes are added at once:

- Database: it keeps dependencies and imports information, now it is small, but it will grow quite a lot (maybe too much? on future pull requests)
- Package: serves as a middle ground between the PackageMetadata and the Database classes, maybe it is too small that should be merged somewhere? I still think it makes sense to keep it, small classes are usually nicer.

@reinout @mauritsvanrees I'm eagerly waiting your feedback 🙂

### DIFF
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+
+class ImportsDatabase(object):
+    """Store all imports and requirements of a package
+
+    Use that information to extract useful data out of it,
+    like requirements unused, etc.
+    """
+
+    def __init__(self):
+        self._requirements = set()
+
+    def add_requirements(self, requirements):
+        self._requirements = set([
+            requirement for requirement in requirements
+        ])
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+import logging
+
+
+logger = logging.getLogger(__name__)
 
 
 class ImportsDatabase(object):
@@ -10,8 +14,32 @@ class ImportsDatabase(object):
 
     def __init__(self):
         self._requirements = set()
+        self._extras_requirements = {}
 
     def add_requirements(self, requirements):
         self._requirements = set([
             requirement for requirement in requirements
         ])
+
+    def add_extra_requirements(self, extra_name, dotted_names):
+        # A bit of extra work needs to be done as pkg_resources API returns
+        # all common requirements as well as the extras when asking for an
+        # extra requirements.
+        only_extra_dotted_names = self._filter_duplicates(dotted_names)
+        if extra_name in self._extras_requirements.keys():
+            self._extras_requirements[extra_name].update(
+                only_extra_dotted_names,
+            )
+
+            logger.warn(
+                'extra requirement "%s" is declared twice on setup.py',
+                extra_name,
+            )
+        else:
+            self._extras_requirements[extra_name] = only_extra_dotted_names
+
+    def _filter_duplicates(self, imports):
+        """Return all items in imports that are not found in reference"""
+        all_imports = set([dotted_name for dotted_name in imports])
+        filtered = all_imports - self._requirements
+        return filtered
--- a/z3c/dependencychecker/package.py
+++ b/z3c/dependencychecker/package.py
@@ -196,3 +196,10 @@ class Package(object):
         self.imports.add_requirements(
             self.metadata.get_required_dependencies()
         )
+
+    def set_declared_extras_dependencies(self):
+        """Add this packages' extras dependencies defined in setup.py to the
+        database
+        """
+        for extra, dotted_names in self.metadata.get_extras_dependencies():
+            self.imports.add_extra_requirements(extra, dotted_names)
--- a/z3c/dependencychecker/package.py
+++ b/z3c/dependencychecker/package.py
@@ -176,3 +176,14 @@ class PackageMetadata(object):
             os.sep,
         )
         sys.exit(1)
+
+
+class Package(object):
+    """The python package that is being analyzed
+
+    This class itself does not much per se, but connects the PackageMetadata
+    with the ImportsDatabase, where the important bits are.
+    """
+
+    def __init__(self, path):
+        self.metadata = PackageMetadata(path)
--- a/z3c/dependencychecker/package.py
+++ b/z3c/dependencychecker/package.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from z3c.dependencychecker.db import ImportsDatabase
 from z3c.dependencychecker.dotted_name import DottedName
 from z3c.dependencychecker.utils import change_dir
 import glob
@@ -187,3 +188,11 @@ class Package(object):
 
     def __init__(self, path):
         self.metadata = PackageMetadata(path)
+        self.imports = ImportsDatabase()
+
+    def set_declared_dependencies(self):
+        """Add this packages' dependencies defined in setup.py to the database
+        """
+        self.imports.add_requirements(
+            self.metadata.get_required_dependencies()
+        )
--- a/z3c/dependencychecker/tests/test_imports_database.py
+++ b/z3c/dependencychecker/tests/test_imports_database.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 from z3c.dependencychecker.dotted_name import DottedName
 from z3c.dependencychecker.db import ImportsDatabase
+from z3c.dependencychecker.tests.utils import get_extras_requirements_names
 from z3c.dependencychecker.tests.utils import get_requirements_names
+from z3c.dependencychecker.tests.utils import get_requirements_names_for_extra
 
 
 def test_no_dependencies():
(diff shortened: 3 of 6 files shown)


z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ImportsDatabase(object):
+
+    def __init__(self):
+        self._requirements = []
+        self._extras_requirements = {}
+
+    def set_requirements(self, requirements_list):
```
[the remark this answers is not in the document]
reinout: Shouldn't it be `add_requirements()`? That is, you're adding items to `_self._requirements`. Sounds like it can happen more than once?
OTOH, that isn't really logical.
But.... then it would make much more sense to say something like `self._requirements = set(requirements_list)`.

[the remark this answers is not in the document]
reinout: Normally, I dislike names with "list" or "dict" in them. Sometimes it has its uses, but here it seems a bit odd.

z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
...
+                    requirement.name,
+                )
+            else:
+                self._requirements.append(requirement)
+
+    def _requirement_already_exists(self, requirement):
+        names = self.get_requirements_names()
+        return bool(requirement.name in names)
+
+    def get_requirements_names(self):
```
reinout: This is one of the points where our coding tastes differ. And thus one of the points where I'm getting a bit philosophical: "do I really dislike this? could I use this? is this handy?"

I've always considered `get_xyz()` and `set_xyz()` methods to be a java naming convention. Instead of `.get_xyz()`, the python style is to do just `.xyz()`. And setting something is "just" assigning something to it.

So.... this isn't a request to change your code, it is more of an observation. I'm currently evaluating whether it might actually *help* sometimes to name getters and setters explicitly.

z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
...
+        return [
+            x.name
+            for x in self._requirements
+        ]
+
+    def set_extra_requirements(self, extra_name, dotted_names):
+        """A bit of extra work needs to be done as pkg_resources API returns
```
reinout: This should be a comment instead of a docstring. Or the docstring needs a clear one-liner at the top.

z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
...
+            self._extras_requirements[extra_name] += new_dotted_names
+
+            logger.warn(
+                'extra requirement "%s" is declared twice on setup.py',
+                extra_name,
+            )
+        else:
+            self._extras_requirements[extra_name] = only_extra_dotted_names
+
+    def _remove_duplicates(self, imports, other=None):
```
reinout: What does this do? Removing duplicates can be done via python sets.

The method name is also a bit confusing. "remove duplicates" sounds like it removes duplicates from some existing variable.
What you seem to be doing is more to filter an input variable and return a filtered list without duplicates.
Something like `._cleaned_requirements()` sounds better to me.

reinout: Similarly: what is `other`? Do you know a clearer name for that?

And if not set, it is the list of the regular requirements' names. But shouldn't 'other' be appended to that regular list? The name doesn't make that clear.

z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
...
+        other_names = other
+        if other is None:
+            other_names = self.get_requirements_names()
+
+        filtered_list = [
+            x
```
reinout: Is it clearer to use `requirement` instead of `x`? You're calling `.name` on it and that's something you can call on requirements, right?
With a proper name I wouldn't have to guess :-)

z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
...
+            for x in imports
+            if x.name not in other_names
+        ]
+
+        return filtered_list
+
+    def _only_new_requirements(self, extra_name, dotted_names):
+        current_names = self.get_requirements_names_for_extra(extra_name)
+        only_new = self._remove_duplicates(
```
reinout: Could you do this with a python set? Duplication handling, subtracting sets, it is all there.

z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
...
+            dotted_names,
+            other=current_names,
+        )
+        return only_new
+
+    def get_extras(self):
```
reinout: In this case, the "get" name really confuses me a bit when reading the code. It sounds like you're downloading some extra stuff... It is probably the combination with "extra" that is the problem.

Do you need this one-liner function? Perhaps you can just say `self._extras_requirements.keys()` in the one or two places where you need it?

z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
...
+        return self._extras_requirements.keys()
+
+    def get_requirements_names_for_extra(self, extra_name):
+        if extra_name not in self.get_extras():
+            return []
+
+        return [
+            x.name
+            for x in self._extras_requirements[extra_name]
```
reinout: With the following line you wouldn't even need the `self.get_extras()` call:

    for x in self._extras_requirements.get(extra_name, [])

reinout: And again use `requirement` instead of `x`

z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
+# -*- coding: utf-8 -*-
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class ImportsDatabase(object):
```
reinout: Add a short doctring to explain the use of this class.

z3c/dependencychecker/package.py
```diff
@@ -176,3 +177,21 @@ def _get_sources_top_level(self):
             os.sep,
         )
         sys.exit(1)
+
+
+class Package(object):
```
reinout: Can you add a short docstring? I think you mean "something like zest.releaser or z3c.dependencychecker" with "package". But better make it explicit.

z3c/dependencychecker/package.py
```diff
@@ -176,3 +177,21 @@ def _get_sources_top_level(self):
...
+
+    def __init__(self, path):
```
reinout: Can you make "path" a bit more explicit? Absolute base path to the root of the package, I guess?

z3c/dependencychecker/package.py
```diff
@@ -176,3 +177,21 @@ def _get_sources_top_level(self):
...
+        self.metadata = PackageMetadata(path)
+        self.imports = ImportsDatabase()
+
+    def get_declared_dependencies(self):
+        """Get this packages' dependencies defined in setup.py"""
+        self.imports.set_requirements(
+            self.metadata.get_required_dependencies()
+        )
+
+    def get_declared_extras_dependencies(self):
```
reinout: This is a getter, but you're calling a setter inside it? (set_extra_requirements)?
Looks weird (or looks like a typo).

z3c/dependencychecker/tests/test_imports_database.py
```diff
@@ -0,0 +1,130 @@
...
+        DottedName('two'),
+        DottedName('three'),
+        DottedName('one'),
+    ])
+
+    last_log = caplog.records[-1]
+    assert '"one" is declared twice on setup.py' == last_log.message
+
+
+def test_declared_dependencies_no_duplicates():
```
reinout: You're testing without duplicates? I see two "one" items in the test.

z3c/dependencychecker/db.py
gforcada: done

z3c/dependencychecker/db.py
gforcada: That's probably due to me reading the excellent [refactoring book](https://martinfowler.com/books/refactoring.html) where the author puts a lot of emphasis in isolating the code and not leaking implementation details where they do not belong (he goes as far as to say that methods should have at maximum 3 statements, no more, and if you can, less if possible.

He's reasoning is that all statements in a method should have the same level of specificity, i.e.

```
self.get_numbers()
self.calculate_median()
self.print_report()
```

In his opinion, even if calculate_median is just one or two statements long, it should still be in a separate method rather than intermixed with highlevel code.

z3c/dependencychecker/db.py
gforcada: As I mentioned above on other comments I (ab)use(?) iterators, so they need to be consumed.

Now I'm wondering if for these specific cases, where the total number of elements (here talking about setup.py requirements and extra requirements) should I rather just return the whole list rather than passing iterators around.

I looks fancier, but to not much gain (at least in this specific code), on modules.py I would keep the iterators, as there can be potentially super big packages with hundreds of files and files with hundreds of imports on them.

I'm keeping them by now, but while fixing all your feedback, if I grew tired of it I might end up changing it.

z3c/dependencychecker/db.py
gforcada: As I noticed that everything is a DottedName, there is no need to use ``.name`` on them, the dunder methods can already take care of it.

But you are right, I renamed x for ``dotted_name``

z3c/dependencychecker/db.py
gforcada: I refactored the whole method:
- added a docstring
- renamed it to ``_filter_duplicates``
- rename the argument ``other`` to ``reference``
- actually we can remove the ``get_requirements_names`` as everything is a DottedName, there is no need to be passing around plain strings, as DottedNames already have all the dunder methods needed

Rather than cleaning I decided to use ``filter`` as it is really a matter of getting the items from "imports" that are not found in "other/reference"

Thanks for pointing things out so that I have to have a second look at the code, just with what I reviewed from your comments I could remove at least one or two methods and make the others far simpler! 🙇

z3c/dependencychecker/db.py
gforcada: Changed to a comment and as we renamed set_requirements to add_requirements I'm doing the same here (add_extra_requirements)

z3c/dependencychecker/db.py
gforcada: The important part here is that we are returning a list of plain strings rather than ``DottedName`` objects, which is what self._requirements is storing. Now that you mention it, yes ``requirement_names`` sounds better, I will change it.

z3c/dependencychecker/db.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __init__(self):
+        self._requirements = []
+        self._extras_requirements = {}
+
+    def set_requirements(self, requirements_list):
+        for requirement in requirements_list:
+            if self._requirement_already_exists(requirement):
+                logger.warn(
+                    '"%s" is declared twice on setup.py',
```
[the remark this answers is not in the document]
gforcada: Oops, my English needs some improvements 😊 , thought this is gone anyway. Still, thanks for pointing it out!

z3c/dependencychecker/db.py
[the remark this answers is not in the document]
gforcada: Actually requirements_list is quite misleading because, actually is an iterator, so I changed that to a plain requirements.

I thought that I did the extra effort to mimic the current z3c.dependencychecker implementation, but a quick search seems that it does not have it, so I will remove it.

To not block the flow of pull requests, if that's ok with you (there are at maximum 2 more to come), I will do a full switch to properties (or cached properties actually) after all code is merge, so that we keep everything consistent style wise (or at least as much consistent as I achieved to be).

I renamed the method to ``add_requirements`` but as mentioned above, we should move that to a property later on.

z3c/dependencychecker/db.py
gforcada: Oh that's totally right, actually with that I could remove two methods already, I never remember the ``get`` method of dictionaries 😅

z3c/dependencychecker/package.py
gforcada: Sure, no problem

z3c/dependencychecker/package.py
gforcada: Right, I renamed both methods to ``set_xxxx``

z3c/dependencychecker/tests/test_imports_database.py
gforcada: Oh, I see, with this test I meant "check that duplicates are removed", so that there are no duplicates after processing them. I will add a docstring on it.

gforcada:
@reinout thanks for the review! 🙇 I did all the changes locally and I'm about to apply all of them and refactor the commits so they still look in place.

gforcada:
@reinout changes made and pushed, if you would have time to do a second review, it will be greatly appreciated

gforcada:
@reinout when refactoring the code to adapt to your review I noticed that those methods on ImportsDatabase that were returning the requirements names, or the extras, or the names for a given extra, were mostly used in tests, so it was ok to remove them from imports_database.py, but then I looked at the tests and I did not like at all to have to access private variables of classes, so I *recreated* those methods in ``tests/utils.py`` so now the tests are nicer to read.

I hope it is ok with you.

reinout:
Looks fine now, you can merge it!
I’ll be off this weekend, so I won’t be able to review a new PR till Sunday afternoon, btw. 

Reinout
(Verstuurd vanaf iPhone, dus kort)

gforcada:
Thanks! PR merged 😃

Enjoy your weekend!

Issue #41, opened 2015-10-27
Allow configuring dependencychecker, especially meta dependencies


mauritsvanrees:
What I mean is: in Plone packages you often do not import `Zope2`, but you import for example `Products.Five`.  But `Products.Five` is included in the `Zope2` package, at least for versions currently in use by Plone.

Let me give an example. For `collective.searchandreplace` I have updated the `setup.py` using dependency checker.  See https://github.com/collective/collective.searchandreplace/blob/663b9c5f7d44ce2f022bed18f4decd02f0a8ba4f/setup.py

Result:

```
Missing requirements
====================
     Products.Five.browser.pagetemplatefile.ViewPageTemplateFile

Missing test requirements
=========================
     Products.Five.fiveconfigure
     Products.Five.testbrowser.Browser
     Testing.ZopeTestCase
     Testing.ZopeTestCase.FunctionalDocFileSuite

Unneeded requirements
=====================
     Products.CMFPlone
     Zope2

```

Those missing requirements are already included in Zope2.  It would be nice if there was a way to tell this to dependencychecker.

Also, `Products.CMFPlone` is not itself imported, but I have added it because the package only makes sense in the context of a Plone site.

So I wonder if there is already something that handles this, or if the code can be improved to somehow realise that those packages are really part of a different package.

Otherwise I wonder if a config file can help.  Say a `~/.dependencychecker.cfg`:

```
[dependencychecker]
# Do not complain about these unneeded requirements:
unneeded_whitelist = Products.CMFPlone
# Maybe also:
# missing_whitelist = ...
# unneeded_test_whitelist = ...
# missing_test_whitelist = ...

[provides]
# Packages that provide (contain) other packages:
Zope2 =
    Products.Five
    Testing
```

reinout:
I like the "provides" solution! That does exactly what's needed.

Would it make sense to put a `[dependencychecker_provides]` section in the `setup.cfg`?

mauritsvanrees:
That makes sense. We can look for it there and in pypirc, just like zest.releaser does.

idgserpro:
Some progress here?

idgserpro:
Maybe you should have a metapackages configuration, instead of having the provides. In case of Zope2 any folder inside the egg should be considered a package and be ignored by dependencychecker.

The configuration would look something like:

```
[dependencychecker]
# Do not complain about these unneeded requirements:
unneeded_whitelist = Products.CMFPlone
# Maybe also:
[...]

PR #64, opened 2017-12-16
Reports

gforcada:
And finally, the previous to last pull request.

After adding a DottedName class #60 , PackageMetadata #61 , modules to extract imports #62 and a Package and Database #63 finally we get some use of it all.

### DIFF
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -78,6 +78,22 @@ class ImportsDatabase(object):
         unique_imports = self._get_unique_imports(imports_list=missing)
         return unique_imports
 
+    def get_unneeded_requirements(self):
+        all_but_test_requirements = self._requirements.copy()
+        for extra in self._extras_requirements:
+            all_but_test_requirements.update(self._extras_requirements[extra])
+        test_requirements = self._get_test_extra()
+        for dotted_name in test_requirements:
+            all_but_test_requirements.remove(dotted_name)
+        filters = (
+            self._filter_out_known_packages,
+            self._filter_out_python_standard_library,
+            self._filter_out_used_imports,
+        )
+        unneeded = self._apply_filters(all_but_test_requirements, filters)
+        unique_imports = self._get_unique_imports(imports_list=unneeded)
+        return unique_imports
+
     @staticmethod
     def _apply_filters(objects, filters):
         """Filter a list of given objects through a list of given filters
@@ -111,6 +127,12 @@ class ImportsDatabase(object):
         unique_dotted_names = list(set(imports_list))
         sorted_unique_dotted_names = sorted(unique_dotted_names)
         return sorted_unique_dotted_names
+
+    def _filter_out_used_imports(self, dotted_name):
+        return self._discard_if_found_obj_in_list(
+            dotted_name,
+            self.imports_used,
+        )
 
     def _filter_out_known_packages(self, dotted_name):
         to_ignore = (
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -15,6 +15,7 @@ class ImportsDatabase(object):
     def __init__(self):
         self._requirements = set()
         self._extras_requirements = {}
+        self.imports_used = []
 
     def add_requirements(self, requirements):
         self._requirements = set([
@@ -43,3 +44,7 @@ class ImportsDatabase(object):
         all_imports = set([dotted_name for dotted_name in imports])
         filtered = all_imports - self._requirements
         return filtered
+
+    def add_imports(self, imports):
+        for single_import in imports:
+            self.imports_used.append(single_import)
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -65,6 +65,19 @@ class ImportsDatabase(object):
         unique_imports = self._get_unique_imports(imports_list=missing)
         return unique_imports
 
+    def get_missing_test_imports(self):
+        filters = (
+            self._filter_out_known_packages,
+            self._filter_out_only_testing_imports,
+            self._filter_out_own_package,
+            self._filter_out_requirements,
+            self._filter_out_test_requirements,
+            self._filter_out_python_standard_library,
+        )
+        missing = self._apply_filters(self.imports_used, filters)
+        unique_imports = self._get_unique_imports(imports_list=missing)
+        return unique_imports
+
     @staticmethod
     def _apply_filters(objects, filters):
         """Filter a list of given objects through a list of given filters
@@ -111,6 +124,10 @@ class ImportsDatabase(object):
     def _filter_out_testing_imports(dotted_name):
         return not dotted_name.is_test
 
+    @staticmethod
+    def _filter_out_only_testing_imports(dotted_name):
+        return dotted_name.is_test
+
     def _filter_out_own_package(self, dotted_name):
         return dotted_name not in self.own_dotted_name
 
@@ -118,6 +135,16 @@ class ImportsDatabase(object):
         return self._discard_if_found_obj_in_list(
             dotted_name,
             self._requirements,
+        )
+
+    def _filter_out_test_requirements(self, dotted_name):
+        test_requirements = self._get_test_extra()
+        if not test_requirements:
+            return True
+
+        return self._discard_if_found_obj_in_list(
+            dotted_name,
+            test_requirements,
         )
 
     def _filter_out_python_standard_library(self, dotted_name):
@@ -163,3 +190,11 @@ class ImportsDatabase(object):
 
         fake_std_libraries = [DottedName(x) for x in libraries]
         return fake_std_libraries
+
+    def _get_test_extra(self):
+        candidates = ('test', 'tests', )
+        for candidate in candidates:
+            if candidate in self._extras_requirements:
+                return self._extras_requirements[candidate]
+
+        return []
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -39,7 +39,7 @@ class ImportsDatabase(object):
             self._extras_requirements[extra_name] = only_extra_dotted_names
 
     def _filter_duplicates(self, imports):
-        """Return all items in imports that are not found in reference"""
+        """Return all items in imports that are not a requirement already"""
         all_imports = set([dotted_name for dotted_name in imports])
         filtered = all_imports - self._requirements
         return filtered
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -94,6 +94,20 @@ class ImportsDatabase(object):
         unique_imports = self._get_unique_imports(imports_list=unneeded)
         return unique_imports
 
+    def get_unneeded_test_requirements(self):
+        test_requirements = self._get_test_extra()
+        if not test_requirements:
+            return []
+
+        filters = (
+            self._filter_out_known_packages,
+            self._filter_out_python_standard_library,
+            self._filter_out_used_imports,
+        )
+        unneeded = self._apply_filters(test_requirements, filters)
+        unique_imports = self._get_unique_imports(imports_list=unneeded)
+        return unique_imports
+
     @staticmethod
     def _apply_filters(objects, filters):
         """Filter a list of given objects through a list of given filters
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
+from stdlib_list import stdlib_list
+from z3c.dependencychecker.dotted_name import DottedName
 import logging
+import sys
 
 
 logger = logging.getLogger(__name__)
@@ -16,6 +19,7 @@ class ImportsDatabase(object):
         self._requirements = set()
         self._extras_requirements = {}
         self.imports_used = []
+        self.own_dotted_name = None
 
     def add_requirements(self, requirements):
         self._requirements = set([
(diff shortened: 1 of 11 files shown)


z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+
+    def get_missing_imports(self):
+        filters = (
+            self._filter_known_packages,
+            self._filter_testing_imports,
+            self._filter_own_package,
+            self._filter_requirements,
+            self._filter_python_standard_library,
+        )
+        filtered = self._process_pipeline(self.imports_used, filters)
```
reinout: Perhaps rename "filtered" to "missing"? You "filter out" stdlib imports, known packages, etcetera. So what's left isn't "filtered", but "remaining". (Or in this case, "missing", I guess).

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+    def get_missing_test_imports(self):
+        filters = (
+            self._filter_known_packages,
+            self._filter_only_testing_imports,
+            self._filter_own_package,
+            self._filter_requirements,
+            self._filter_test_requirements,
+            self._filter_python_standard_library,
+        )
+        filtered = self._process_pipeline(self.imports_used, filters)
```
reinout: Also here, "filtered" => "missing"

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+        test_requirements = self._get_test_extra()
+        if test_requirements:
+            for dotted_name in test_requirements:
+                all_but_test_requirements.remove(dotted_name)
+        filters = (
+            self._filter_known_packages,
+            self._filter_python_standard_library,
+            self._filter_used_imports,
+        )
+        unneeded = self._process_pipeline(all_but_test_requirements, filters)
```
reinout: Here the name is clear (as it isn't 'filtered').

Looks like a nice, clean approach: those filters and `self._process_pipeline()`!

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+            self._filter_testing_imports,
+            self._filter_python_standard_library,
+            self._filter_own_package,
+        )
+        non_testing_imports = self._process_pipeline(
+            self.imports_used,
+            filters,
+        )
+        requirements_not_used = [
+            x
```
reinout: x => requirement

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+            for x in self._requirements
+            if self._filter_requirements_lists(non_testing_imports, x)
+        ]
+        filters = (
```
reinout: You're overwriting an earlier variable. In cases like that, I've often found it less error-prone to explicitly call them `filters1`, `filters2`. 
I tend to make mistakes with code that re-uses the same name (about once every year).

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+            self._filter_only_testing_imports,
+            self._filter_python_standard_library,
+            self._filter_own_package,
+        )
+        testing_imports = self._process_pipeline(
+            self.imports_used,
+            filters,
+        )
+        should_be_test_requirements = [
+            x
```
reinout: x => requirement

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+            self._filter_known_packages,
+            self._filter_python_standard_library,
+            self._filter_used_imports,
+        )
+        unneeded = self._process_pipeline(test_requirements, filters)
+        unique_imports = self._get_unique_imports(imports_list=unneeded)
+        return unique_imports
+
+    def _process_pipeline(self, objects, filters):
+        final_list = []
```
reinout: Rename final_list => result?

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+        filters = (
+            self._filter_known_packages,
+            self._filter_python_standard_library,
+            self._filter_used_imports,
+        )
+        unneeded = self._process_pipeline(test_requirements, filters)
+        unique_imports = self._get_unique_imports(imports_list=unneeded)
+        return unique_imports
+
+    def _process_pipeline(self, objects, filters):
```
reinout: This needs a quick explanation of what you're doing here.
A filter function should apparently return False or True. What is the meaning when it returns "True"? Does the item gets filtered out or does it get included?

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+            else:
+                final_list.append(single_object)
+
+        return final_list
+
+    def _get_unique_imports(self, imports_list=None):
+        if imports_list is None:
+            imports_list = self.imports_used
+        unique_dotted_names = list(set(imports_list))
+        sorted_unique_dotted_names = sorted(unique_dotted_names)
```
reinout: Sorting them is neat and tidy!

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+        return sorted_unique_dotted_names
+
+    def _filter_used_imports(self, dotted_name):
```
reinout: What do you think about renaming those methods from `_filter_something()` to `_filter_out_something()`?

The methods return True or False. Having a "negative" method name makes it harder to interpret "False". "do_not_include=False", what does that mean?

Hm. I also see `_filter_testing_imports()` that returns True/False when the dotted name is a test or not. But `_filter_known_packages()` returns a list of requirements.

=> **Something** has to be changed regarding the naming, I think?

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+            test_requirements,
+            dotted_name,
+        )
+
+    def _filter_python_standard_library(self, dotted_names):
+        std_library = self._build_std_library()
+        return self._filter_requirements_lists(std_library, dotted_names)
+
+    @staticmethod
+    def _filter_requirements_lists(requirements, candidate):
```
reinout: Make it clearer what happens here. 

- I see the word "candidate" for the first time here.

- Assuming "r" is a single "requirement", what is the meaning of "candidate in requirement"? I'd expect "candidate == requirement"? But that's weird, too.

- "return not any" is not something I can figure out right away. Something with double negatives again, I guess :-) Using an intermediary variable name should make it clearer.

z3c/dependencychecker/db.py
```diff
@@ -43,3 +48,186 @@ def _filter_duplicates(self, imports):
...
+        return not any((r for r in requirements if candidate in r))
+
+    @staticmethod
+    def _build_std_library():
+        python_version = sys.version_info
```
reinout: This makes the result of the dependencychecker output dependent on the python version, but that is not something that can be helped.
Perhaps we should add a note in the README later on.
OTOH, it ought to be clear enough for people using this tool, right?

z3c/dependencychecker/report.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+
+
+class Report(object):
+
+    def __init__(self, package):
+        self._database = package.imports
+
+    def __call__(self):
```
reinout: I haven't seen yet how you call this class, but a regular method name *might* be clearer.
Not sure yet, I'm just writing it down as I go through the file.

reinout:
The code as a whole look nicely clean and well-thought-ought, as usual!

z3c/dependencychecker/db.py
```diff
@@ -15,6 +18,8 @@ class ImportsDatabase(object):
     def __init__(self):
         self._requirements = set()
         self._extras_requirements = {}
+        self.imports_used = []
+        self.package_name = ''
```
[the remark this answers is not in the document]
gforcada: Done.

z3c/dependencychecker/db.py
gforcada: Done

z3c/dependencychecker/db.py
gforcada: Done

z3c/dependencychecker/db.py
gforcada: Thanks 😊

z3c/dependencychecker/db.py
gforcada: Done (as you have seen, usually I tend to use ``x`` on list comprehensions only where the value on each iteration can be easily spotted, but still, I agree that putting a name is always better, I'm lazy on list comprehensions 🙂 )

z3c/dependencychecker/db.py
gforcada: Done

z3c/dependencychecker/db.py
gforcada: Done

z3c/dependencychecker/db.py
gforcada: Done

z3c/dependencychecker/db.py
gforcada: And makes tests easier to reproduce and assert 🙂

z3c/dependencychecker/db.py
gforcada: Done

z3c/dependencychecker/db.py
gforcada: ``_filter_requirements_lists`` is a bit involved, but all ``_filter_XXX`` methods do return either ``True`` or ``False`` (regarding ``_filter_testing_imports`` and ``_filter_known_packages``).

Regarding renaming to ``filter_out``, sure no problem.

z3c/dependencychecker/db.py
gforcada: I had to battle with this one for quite a bit, let's see if the changes I made make it better:
- I renamed it to ``_discard_if_found_obj_in_list`` and gets two parameters, the object (a ``DottedName``) and a list (of ``DottedName``s)
- I renamed ``candidate`` to ``obj``

Regarding the logic.

I'm using a few tricks of generators:
- ``any`` consumes an iterator and returns True as soon as one element on the iterable is True (see [the official docs](https://docs.python.org/3/library/functions.html#any))
- with that in mind, we pass a generator to ``any`` that tests if the object is in the requirement we are iterating over, as it is a generator, we don't need to go through all possible 300k objects in the list to return, but as soon as one returns True, we are done.
- ``not any`` is because we want to return ``True`` when the object *is found*
- I added plenty of comments and a docstring

z3c/dependencychecker/db.py
gforcada: Now that you mention it... good question!

Given that z3c.dependencychecker can be ``pip install``ed, one would expect that it uses the same python version as the rest of the code installed on that virtual environment...

**But** what about code that tries to be compatible with Python 2 and Python 3? On Python 3 there's quite a few new/removed modules on the standard library, so when running on py2 you will get warnings for the new ones, and the other way around...

Should we use #41 as a way to signal which versions of python standard library to check against, and fallback to the current python version?

z3c/dependencychecker/report.py
gforcada: This is how it is used: https://github.com/reinout/z3c.dependencychecker/blob/gforcada-rewrite/z3c/dependencychecker/main.py#L22-L23

I'm leaving it as is, but if you want, we can change it for a name, no problem.

gforcada:
@reinout I finally got around to finish implementing all your requested changes and suggestions, thanks for the thoughtful review!

Let me know if there's further edits that I should made.

gforcada:
*Sorry I closed it by accident*

z3c/dependencychecker/db.py
reinout: I haven't used `any` yet. Sounds quite useful. I must admit to not using generators as much as I probably should :-)

But: "any returns True as soon as one element is True" plus "`not any` because we want to return True if found" => ??? I'll have to look at the implementation again, I guess, because this sounds like you return False if anything is found.

z3c/dependencychecker/db.py
reinout: I think we don't have to do anything fancy, after all. Dependencychecker is aimed at a quite advanced audience, so they'll know it when they see an anomaly from a python2/3 difference. Perhaps a line in the readme, that will be enough.

reinout: An alternative would be to just add the 2.7 and 3.6 dependencies in one big list. That way we'll even handle "except ImportError" conditional imports to handle 2/3 issues. Would that be an idea?

z3c/dependencychecker/report.py
reinout: It is the only place in the code where `__call__()` is used in this way, so a name is clearer. `.print_report()` or `.write_output()` or something like it?

z3c/dependencychecker/db.py
```diff
@@ -39,7 +44,234 @@ def add_extra_requirements(self, extra_name, dotted_names):
...
+            self._filter_out_known_packages,
+            self._filter_out_python_standard_library,
+            self._filter_out_used_imports,
+        )
+        unneeded = self._process_pipeline(test_requirements, filters)
+        unique_imports = self._get_unique_imports(imports_list=unneeded)
+        return unique_imports
+
+    @staticmethod
+    def _process_pipeline(objects, filters):
```
reinout: I keep finding it hard to read this part of the code. I've been thinking about it for more than an hour now, trying to come up with a small naming change or so that might clear it up. But I just cannot figure out nice clear names.

I think there are two possible reasons:

a) Double or triple negatives. "filter_out" is negative. "if not" is negative. "discard from" is negative. "if not filtered out" is a double negative. "if not filtered out.... else..." is *another* step. Double negatives are hard to reason about (and thus it is easy to make a mistake with them).

b) I think you're trying to hard. Breaking loops early to make it efficient. Splitting everything into small functions. The `discard if obj found in list` re-implementation of basic comparison functions to make it more efficient. Aren't you re-implementing python's build-in and very efficient `set` operators?

Is it really necessary? There won't be so many imports that stuff won't fit into a list, I guess. And if you're going to go through every list/generator multiple times, you can just turn it into a list. Or a set. Most of the `_process_pipelines()`-using functions can be rewritten something like this:

    test_requirements = self._get_test_extra()
    unneeded = set(test_requirements) - set(known_packages) - set(standard_lib) - ...
    return sorted(unique_imports)

At least, that's what it looks like to me. It is also much more explicit. A "bit" more brute-force. But I don't mind if running dependencychecker takes 5 seconds or 2 seconds :-)

OTOH:

- Perhaps it is needed for what comes in the last PR.
- The code works, we can always refactor it later.

But: I'm a bit afraid of the complexity.

z3c/dependencychecker/report.py
gforcada: Done! I went for ``.print_report()``

z3c/dependencychecker/db.py
```diff
@@ -39,7 +44,234 @@ def add_extra_requirements(self, extra_name, dotted_names):
...
+        missing = self._process_pipeline(self.imports_used, filters)
+        unique_imports = self._get_unique_imports(imports_list=missing)
+        return unique_imports
+
+    def get_unneeded_requirements(self):
+        all_but_test_requirements = self._requirements.copy()
+        for extra in self._extras_requirements:
+            all_but_test_requirements.update(self._extras_requirements[extra])
+        test_requirements = self._get_test_extra()
+        if test_requirements:
```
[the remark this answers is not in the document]
gforcada: Done! Actually that method ``._get_test_extra()`` was returning ``False`` if there was no extra, but I changed that to an empty list, so it will always return a list, which makes it easier.

z3c/dependencychecker/db.py
gforcada: I get your point and I see that I totally missed the ``set`` builtin, shame on me, but at this late on the game, I would rather prefer to keep it as it is (I can rename it to ``apply_filters`` of course) and refactor it later to use sets, would that be ok with you?

The more it goes, the more I'm interested in getting a version out in pypi and python 3 compatible 🙂

z3c/dependencychecker/db.py
```diff
@@ -39,7 +44,234 @@ def add_extra_requirements(self, extra_name, dotted_names):
...
+            dotted_name,
+            test_requirements,
+        )
+
+    def _filter_out_python_standard_library(self, dotted_name):
+        std_library = self._build_std_library()
+        return self._discard_if_found_obj_in_list(dotted_name, std_library)
+
+    @staticmethod
+    def _discard_if_found_obj_in_list(obj, obj_list):
```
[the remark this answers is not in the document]
gforcada: If you could be forgetful 🙇  about this method, until we refactor it out with sets, I that's ok with you.

z3c/dependencychecker/db.py
reinout: Valid point :)
Then we'll refactor it to sets afterwards. We can fix the method name then, too. I'll just merge the pull request :)

PR #67, opened 2018-01-02
New implementation

gforcada:
As all the pieces are already in master, it is time to say goodbye to the old implementation and say hello to the new one 😄 

As there is [a problem with zc.buildout and Python +3.5](https://github.com/buildout/buildout/issues/434) the python 3 support is limited to 3.3 and 3.4.


setup.py
```diff
@@ -17,8 +17,24 @@
       dependencies.""",
       long_description=long_description,
       # Get strings from http://www.python.org/pypi?%3Aaction=list_classifiers
-      classifiers=[],
-      keywords=[],
+      classifiers=[
+          'Development Status :: 5 - Production/Stable',
+          'Environment :: Console',
+          'Intended Audience :: Developers',
+          'License :: OSI Approved :: Zope Public License',
```
gforcada: Does it make sense to keep this license with the complete rewrite that I did? There are no longer any parts left that are coming from the old code. I would be happier to change the license to (L)GPLv3 or similar, but I'm not a lawyer, so I have no clue on these things 😅

reinout: Getting rid of the ZPL: good idea. (L)GPL or BSD, all three are fine.

Most common nowadays: BSD, which means that everyone can copy your code around and can even use it in commercial products and change the license and so. GPL is more strict in that it has to stay open source and that software that uses it also has to be GPL. LGPL allows you to use it in non-GPL code.

Normally I'd personally use LGPL, but in the case of this tool I'd say BSD. It is makes it a bit easier to (re-)use it in other tools, which is probably what we want in the end.

gforcada:
@reinout license updated 🙂

CHANGES.rst
```diff
@@ -1,10 +1,13 @@
...
 
-1.17 (unreleased)
------------------
+2.0 (unreleased)
+----------------
 
-- Nothing changed yet.
+- Complete rewrite: code does no longer use deprecated functionality,
+  is more modular, more pythonic, easier to extend and hack, and above all,
+  has a 100% test coverage to ensure that it works as expected.
```
reinout: And it runs on python 3 :-)

setup.py
```diff
@@ -10,35 +10,58 @@
...
+    keywords=[
+        'dependencies',
+        'requirements',
+        'missing',
+        'imports',
+    ],
+    author='Reinout van Rees',
+    author_email='reinout@vanrees.org',
+    url='https://github.com/reinout/z3c.dependencychecker',
+    license='ZPL',
```
reinout: You'll need BSD here, too

z3c/dependencychecker/main.py
```diff
@@ -0,0 +1,80 @@
...
+    if len(args) < 1:
+        logger.debug('path used: %s', path)
+        return path
+
+    path = os.path.abspath(args[0])
+    if os.path.isdir(path):
+        logger.debug('path used: %s', path)
+        return path
+
+    logger.error('Given path does is not a folder: %s', args[0])
```
reinout: Remove 'does'.

z3c/dependencychecker/modules.py
```diff
@@ -308,7 +308,7 @@ def _parse_docstring(self, docstring):
                 try:
                     tree = ast.parse(code)
                 except SyntaxError:
```
reinout: I wonder, should we log this at debug level? Not sure.

z3c/dependencychecker/package.py
```diff
@@ -209,5 +214,7 @@ def set_declared_extras_dependencies(self):
     def analyze_package(self):
         top_folder = self.metadata.top_level
         for module_obj in MODULES:
+            logger.debug(module_obj)
             for source_file in module_obj.create_from_files(top_folder):
+                logger.debug('Search deps on file %s', source_file.path)
```
reinout: "on file" => "in file"?
(And 'deps' => 'dependencies'?)

setup.py
gforcada: Oops, done!

CHANGES.rst
gforcada: Added

z3c/dependencychecker/main.py
gforcada: 😊 oooops

z3c/dependencychecker/package.py
gforcada: Done

z3c/dependencychecker/modules.py
gforcada: Good question... I see that we have a couple of ``except SyntaxError``, so if we do, we should  do it in both.

While debugging myself I noticed this SyntaxError exceptions while working on the python 3 porting. There is a ``    >>> print 7`` on the ``sample1`` test project, which breaks on python 3, is that useful for users of z3c.dependencychecker? I'm not so sure.

As you are suggesting to add it at debug level, it is not going to be that noticeable anyway. I'm adding it. At least for us while debugging is going to be useful.

reinout:
Hurray, I've released version 2.0! Thanks a lot for your work. 
I've learned a couple of new tricks in the process. Like how pytest works, for instance.

reinout:
Now I'm going to try it out on a couple of projects :-)

gforcada:
Oh cool, I would have made a beta if rc, but that's fine as well :-)

I have some fixes piled up as well. I will be sending mire pull requests :-)

PR #69, opened 2018-01-04
Logging output tweaks

reinout:
Some tweaks I did while trying out 2.0 with the `-v` option :-)


z3c/dependencychecker/db.py
```diff
@@ -51,7 +51,7 @@ def _filter_duplicates(self, imports):
 
     def add_imports(self, imports):
         for single_import in imports:
-            logger.debug('Import found %s', single_import.name)
+            logger.debug('    Import found: %s', single_import.name)
```
gforcada: Cool, great idea!

z3c/dependencychecker/main.py
```diff
@@ -55,6 +55,7 @@ def set_log_level(verbose):
     logging.basicConfig(
         level=level,
         stream=sys.stdout,
+        format="%(levelname)s: %(message)s"
```
gforcada: Aha! I knew this could be done, but I never bothered to check it 🙂

z3c/dependencychecker/package.py
```diff
@@ -20,6 +20,7 @@ class PackageMetadata(object):
     """
 
     def __init__(self, path):
+        logger.debug("Reading package metadata for %s...", path)
```
gforcada: I always tried to use single quotes, should we try to keep them consistent across the code? There are flake8 plugins for that as well

z3c/dependencychecker/package.py
```diff
@@ -214,10 +215,11 @@ def set_declared_extras_dependencies(self):
     def analyze_package(self):
         top_folder = self.metadata.top_level
         for module_obj in MODULES:
-            logger.debug(module_obj)
+            logger.debug("Starting analyzing files using %s...", module_obj)
```
gforcada: On the next debug line you are already pointing which ``module_obj`` is being used, so the output now looks like:

```
DEBUG: Package name is plone.api
DEBUG: Starting analyzing files using <class 'z3c.dependencychecker.modules.PythonModule'>...
DEBUG: Searching dependencies (with PythonModule) in file /home/gil/Documents/plone/plone5/plone/src/plone.api/src/plone/__init__.py...
DEBUG: Searching dependencies (with PythonModule) in file /home/gil/Documents/plone/plone5/plone/src/plone.api/src/plone/api/env.py...
DEBUG:     Import found: AccessControl.SecurityManagement.getSecurityManager
```

IMHO this line is superfluous, as the information is already provided on the following lines. What do you think?

OTOH we could add some padding on the ``Search dependencies`` and double the padding on ``Import found`` so it could look like this:

```
DEBUG: Starting analyzing files using 'PythonModule'...
DEBUG:     Searching dependencies in file /home/gil/Documents/plone/plone5/plone/src/plone.api/src/plone/api/env.py...
DEBUG:         Import found: AccessControl.SecurityManagement.getSecurityManager
DEBUG:         Import found: AccessControl.SecurityManagement.getSecurityManager
DEBUG:         Import found: AccessControl.SecurityManagement.getSecurityManager
DEBUG:     Searching dependencies in file /home/gil/Documents/plone/plone5/plone/src/plone.api/src/plone/api/env.py...
DEBUG:         Import found: AccessControl.SecurityManagement.getSecurityManager
DEBUG: Starting analyzing files using 'ANOTHERMODULE'...
DEBUG:     Searching dependencies in file /home/gil/Documents/plone/plone5/plone/src/plone.api/src/plone/api/env.py...
```

In any case, we should change the ``module_obj`` here with ``module_obj.__name__`` as you did below.

z3c/dependencychecker/package.py
```diff
@@ -214,10 +215,11 @@ def set_declared_extras_dependencies(self):
...
             for source_file in module_obj.create_from_files(top_folder):
                 logger.debug(
-                    'Search dependencies in file %s',
+                    'Searching dependencies (with %s) in file %s...',
+                    module_obj.__name__,
```
gforcada: 👍

gforcada:
@reinout there was a test failing (as I was testing the debug output). I took the liberty to rebase the branch on top of master and fix the bug, I hope it is ok with you.

gforcada:
@reinout should we add a change log entry? I got used on personal/work projects, as well as Plone ones, that whenever I make a pull request, there is always a change log entry on CHANGES.rst this way it is easier to keep a highlevel overview of what's going on from version to version without even having to look at the repository, just look on pypi and you are done.

z3c/dependencychecker/package.py
reinout: That's where our styles differ. For short words, I tend to use single quotes. For sentences I **always** use double quotes. In 25% of the sentences you have a word like `can't` or `persons'` and I don't want to backslash-escape single quotes if I can just just double quotes.

I'm pretty sure the original code used double quotes :)

z3c/dependencychecker/package.py
reinout: I don't want to get too clever with indentation. One level in one specific place is bad enough.

The line has use as it shows the flow of the program. I'd keep the long "class" string in that case. The reason: I could imagine us adding a plugin mechanism later on in this exact spot that lets people add custom readers.

reinout:
Regarding changelog: I didn't consider a change in debug logging worth a changelog line. But: it *does* show that we're paying attention to the debug logging. So I'll add a line. Good idea.

reinout:
Done.

PR #70, opened 2018-01-04
Minor fixes

gforcada:
While testing the code with packages I notices quite a few things, but as it was too late to merge them on the *right* pull request I kept them until #67 was merged.

Overview of the changes:
- minor cleanups
- handle ZCML with more than one DottedName on an attribute
- process requirements to strip the ``python-`` prefix as well as to turn dashes into underscores
- mark all DocFiles imports as being test ones, regardless of where they are in the package structure

### DIFF
--- a/z3c/dependencychecker/dotted_name.py
+++ b/z3c/dependencychecker/dotted_name.py
@@ -10,7 +10,6 @@ class DottedName(object):
         self,
         name,
         file_path=None,
-        line_number=None,
         is_test=False,
     ):
         self.name = name
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,14 @@ Changelog of z3c.dependencychecker
 - Small improvements to the debug logging (``-v/--verbose`` shows it).
   [reinout]
 
+- Remove unused parameter in DottedName.
+  [gforcada]
+
+- All imports found by DocFiles imports extractor are marked as test ones.
+  [gforcada]
+
+- Handle multiple dotted names found in a single ZCML parameter.
+  [gforcada]
 
 2.0 (2018-01-04)
 ----------------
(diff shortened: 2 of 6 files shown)


z3c/dependencychecker/dotted_name.py
```diff
@@ -25,8 +25,12 @@ def from_requirement(cls, requirement, file_path=None):
         """A requirement in this method's context is a
         pkg_resources.Requirement
         """
+        # some pypi packages have a python- prefix that is not part of the
```
reinout: Well, this is just one of the many conventions. A django prefix is also quite often used. But sometimes it really ought to keep the django prefix and sometimes it needs to be used without.

Likewise, stripping the python prefix seems very dangerous to me.

reinout: I'm getting the impression that the new implementation treats imports and requirements the same. Both are converted to a single dotted name. So there's no room for:

- Trying to import something to figure out which requirement name would match.
- A package providing multiple top-level paths (like some zope "meta-packages" do).

So I think something a bit more explicit is needed here. A distinction between imports (which are really dotted names) and packages/requirements, which might have to be handled a little bit special.

reinout: In my tests, I've seen a package "pyjwt" that is imported as 'jwt'.

Package `fake-factory` as `import faker`.

django-appconf => appconf

django-oidc-provider => oidc_provider

django-tls => tls

So if you'd also handle 'py' and 'django' in the same way, I'd see less unrecognized packages. But I'm quite afraid of ruining some real python-something or django-something packages.

gforcada:
@reinout I removed the prefix removal and kept all the others, which I think we agree on, I will create a separate pull request for that.

If you are fine with the changes left, you can merge this one as you please.

PR #72, opened 2018-01-06
WIP: pypi names

gforcada:
*Only merge this one after #70 is merged, as it contains its commits as well.*

Continuing the discussion from #70:

When I was looking at that part of the code that tried to import a dotted name and from there find out to which package belonged, I thought that was really clever, but then this means that you should have that package ``pip install``ed in your environment where you are actually running z3c.dependencychecker, which is far from optimal.

IMHO we should give a go at #41 I have a half baked branch where I'm reading pyproject.toml to get the metadependencies and from there override the reports. We could as well let the users specify which dotted names belong to a package.

With this we would allow a completely clean and isolated virtual environment for z3c.dependencychecker, without having to install any extra dependencies.

Would that be ok with you? @reinout @mauritsvanrees

### DIFF
--- a/z3c/dependencychecker/dotted_name.py
+++ b/z3c/dependencychecker/dotted_name.py
@@ -1,6 +1,16 @@
 # -*- coding: utf-8 -*-
 import hashlib
+import re
 from functools import total_ordering
+
+
+# some pypi packages' names have a prefix that is not part of the package
+# namespace
+# They can be safely removed
+PREFIXES = (
+    'python-',
+    'django-',
+)
 
 
 @total_ordering
@@ -24,8 +34,12 @@ class DottedName(object):
         """A requirement in this method's context is a
         pkg_resources.Requirement
         """
+        dotted_name = requirement.project_name
+        for prefix in PREFIXES:
+            dotted_name = re.sub(r'^{0}'.format(prefix), '', dotted_name)
+        dotted_name = dotted_name.replace('-', '_')
         return cls(
-            requirement.project_name,
+            dotted_name,
             file_path=file_path,
         )
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,11 @@ Changelog of z3c.dependencychecker
 - Handle multiple dotted names found in a single ZCML parameter.
   [gforcada]
 
+- Process requirements' project names to remove the ``python-`` prefix,
+  turn all dashes into underscores as well.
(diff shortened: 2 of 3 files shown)


reinout:
Using/advocating `pyproject.toml` for something extra like this? Fine.

Using a mapping in a `pyproject.toml` file for pypi names/imports: fine. Nicely explicit.

What I would like: a separate command line program that would try to import the imports with missing packages and that would report on found imports. Perhaps it can be integrated into the main program as well. 

- We *do* know the missing imports. 
- We *can* try/except import them.
- We *can* report if we've found something.
- ... and then tell the user where (`pyproject.toml`!) to insert lines telling dependencychecker about the requirement/import mapping.

That seems like a happy middle ground?

reinout:
I restarted one of the travis builds; there was a temporary glitch. Everything is green again.

z3c/dependencychecker/dotted_name.py
```diff
@@ -1,8 +1,18 @@
 # -*- coding: utf-8 -*-
 import hashlib
+import re
 from functools import total_ordering
 
 
+# some pypi packages' names have a prefix that is not part of the package
+# namespace
+# They can be safely removed
```
reinout: My point was that they *cannot* be safely removed. Here are some paths from the first random django project I looked at:

 '/code/eggs/django_tls-0.0.2-py3.5.egg',
  '/code/eggs/django_oidc_provider-0.5.2-py3.5.egg',
  '/code/eggs/django_nose-1.4.4-py3.5.egg',
  '/code/eggs/django_extensions-1.6.7-py3.5.egg',
  '/code/eggs/django_appconf-1.0.2-py3.5.egg',

Of those five, three should get the django prefix removed, two (django-extensions and django-nose) should keep their prefix.

reinout:
The prefix *is* very common. But I think we're getting too many false positives.

What we can do is to do the prefix logic after the fact. Look at missing imports (both regular and test/extra) and extraneous requirements. Then it is easy to suggest that "tls" and "django-tls" might have to be mapped. And then we should let the user make a decision about it, whether he puts it in the mapping in the config file.

Something like that?

gforcada:
Sounds like a good plan. Then I will try to finish first the ``pyproject.toml`` support and then we can do this. Without the mapping, it would be pointless.

gforcada:
There's no need to have this anymore.

PR #73, opened 2018-01-15
Use properties

gforcada:
@reinout as you once suggested, using properties would make the code a bit more clear.

Let's see if that's what you were expecting.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -16,6 +16,7 @@ show-picked-versions = true
 setuptools = 33.1.1
 zc.buildout = 2.9.5
 stdlib-list = 0.4.0
+cached-property = 1.3.1
 
 z3c.dependencychecker =
 configparser = 3.5.0
--- a/z3c/dependencychecker/dotted_name.py
+++ b/z3c/dependencychecker/dotted_name.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import hashlib
 from functools import total_ordering
+from cached_property import cached_property
 
 
 @total_ordering
@@ -13,9 +14,8 @@ class DottedName(object):
         is_test=False,
     ):
         self.name = name
-        self.safe_name = self._get_safe_name(name)
-        self.namespaces = self._get_namespaces(self.safe_name)
-        self.is_namespaced = self._is_namespaced(self.namespaces)
+        self.safe_name = name.lower().replace('-', '_')
+
         self.file_path = file_path
         self.is_test = is_test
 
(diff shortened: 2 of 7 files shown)


z3c/dependencychecker/dotted_name.py
```diff
@@ -29,19 +29,18 @@ def from_requirement(cls, requirement, file_path=None):
             file_path=file_path,
         )
 
-    @staticmethod
-    def _get_safe_name(name):
-        safe_name = name.lower().replace('-', '_')
-        return safe_name
+    @cached_property
+    def safe_name(self):
+        self._safe_name = self._safe_name.lower().replace('-', '_')
```
reinout: Setting `self._safe_name` is a bit of side effect (and thus dangerous, as it is only set after the first time you've called `self.safe_name`.

In this case I'd probably put `self.safe_name = .....lower().replace()` directly in the class's init (or keep the original method call, but with a different name).

buildout.cfg
```diff
@@ -16,6 +16,7 @@ show-picked-versions = true
 setuptools = 33.1.1
 zc.buildout = 2.9.5
 stdlib-list = 0.4.0
+cached-property = 1.3.1
```
reinout: In most cases, a regular `@property` would have been enough, btw. 

Such a `@cached_property` is mostly useful when something is computationally intensive (and only on the condition that data doesn't change). So it isn't strictly needed as a dependency.

(I *do* like it, of course).

z3c/dependencychecker/package.py
```diff
@@ -21,32 +22,32 @@ class PackageMetadata(object):
...
+        if 'setup.py' not in os.listdir(self._path):
             logger.error(
                 'setup.py was not found in %s. '
                 'Without it dependencychecker can not work.',
-                path,
+                self._path,
             )
             sys.exit(1)
-        return path
+        self._distribution_root = self._path
```
reinout: Same remark as for "name" previously. Setting _distribution_root is a side effect.

buildout.cfg
gforcada: Given how to code is right now, without the caching, there would be several properties (especially in ``PackageMetadata``) that would be called over and over, so I think is a quite neat solution 👍

z3c/dependencychecker/dotted_name.py
gforcada: Ok, I moved it to the ``__init__``

z3c/dependencychecker/package.py
gforcada: Right, rather than setting a ``self._distribution_root`` we can simply return ``self._path`` to avoid the side effect.

And as we cache the property, it does not matter that we are doing some checks there.

gforcada:
@reinout thanks for the review, I updated the code, I think I addressed your concerns.

reinout:
Travis seems to have problems. If the tests run for you locally, feel free to merge it.

gforcada:
Yes, they do pass locally and coverage remains at 100%

PR #74, opened 2018-01-18
No code analysis on 3.5 and 3.6

gforcada:
That was rather stupid from me 😅 

Tests do pass on 3.5 and 3.6, the only problem there is that code analysis can not be installed, but a z3c.dependencychecker user does not care about plone.recipe.codeanalysis 😄


gforcada:
Wow, that was fast 😄

PR #76, opened 2018-02-11
[performance] Filter packages when adding them

gforcada:
There are some filters that are will be always applied,

so rather than applying them at each report, apply them when they are added to the database.

### DIFF
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -50,17 +50,24 @@ class ImportsDatabase(object):
         return filtered
 
     def add_imports(self, imports):
+        filters = (
+            self._filter_out_known_packages,
+            self._filter_out_own_package,
+            self._filter_out_python_standard_library,
+        )
         for single_import in imports:
             logger.debug('    Import found: %s', single_import.name)
-            self.imports_used.append(single_import)
+
+            unknown_import = self._apply_filters([single_import, ], filters)
+            if unknown_import:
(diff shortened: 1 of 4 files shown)


z3c/dependencychecker/db.py
```diff
@@ -50,30 +50,34 @@ def _filter_duplicates(self, imports):
...
+        filters = (
+            self._filter_out_known_packages,
+            self._filter_out_own_package,
+            self._filter_out_python_standard_library,
+        )
         for single_import in imports:
             logger.debug('    Import found: %s', single_import.name)
-            self.imports_used.append(single_import)
+
+            unknown_import = self._apply_filters([single_import, ], filters)
```
reinout: The term "unknown" might be a bit confusing. After all, what is *more* known than `import os` or `import our_own_package`? :-)

I'm not sure about a correct name. `standard_import`? `already_known_import`? `to_ignore_import`? `to_ignore_because_known_import`?

Does this trigger something?

gforcada: In this context an `unknown_import` is actually the other way around: it is an import that was not filtered, i.e. no `import os`, no own package, etc etc but rather a `from PIL import image`.

Hence the `unknown_import` as the filters applied did not match and thus the import is going to be added on the list of imports used by the package.

reinout: Ah, ok!

PR #77, opened 2018-02-11
Update zc.buildout and setuptools

gforcada:


gforcada:
Unfortunately, it seems that newer zc.buildout and setuptools does not fix the scripts 🙁

reinout:
Those new versions *should* work just fine.
I'm aware of you opening an issue on zc.buildout. I haven't gotten around to debugging it yet.
Ideally, a newer version shouldn't make it worse, right? So in that sense, an undelete of this pull request's branch could help clear things up.

gforcada:
Right, I was mostly interested in seeing if the code analysis problem was solved. I remove that bit and now is a pretty straightforward patch. 🙂

gforcada:
@reinout I hope you did not mind that I merged it.

reinout:
Totally fine

PR #78, opened 2018-03-18
Fix README format

gforcada:
With this, hopefully, pypi displays the description nicely formatted.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,4 @@
+=====================
 z3c.dependencychecker
 =====================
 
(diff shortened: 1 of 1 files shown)


reinout:
I don't think this helps. What I see now on pypi looks like an internal pypi rendering error. The actual header format of z3c.dependencychecker is what I use for most of my open source projects: it works just fine. And it worked just fine for dependencychecker, too, in the past.

gforcada:
So let's see on the next release the hiccup is gone...

PR #79, opened 2018-06-16
Ignore relative imports

gforcada:
It seems that we forgot to ignore those :-)

i.e. imports like `` from . import foo`` or ``from .local import foo``.


reinout:
Good one!

PR #81, opened 2018-06-16
Ignore packages

gforcada:
And with this, I think we have the last building block.

When I've been trying to use it on Plone code bases this was the last missing piece that ``z3c.dependencychecker`` needed to be 100% usable.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,6 +64,29 @@ Add a `pyproject.toml` file on the root of your project with the following conte
 
 z3c.dependencychecker will read this information and use it on its reports.
 
+Ignore packages
+---------------
+
+Sometimes you need to add a package in `setup.py` although you are not importing it directly,
+but maybe is an extra dependency of one of your dependencies,
+or your package has a soft dependency on a package,
+and as a soft dependency it is not mandatory to install it always.
+
+`z3c.dependencychecker` would complain in both cases.
+It would report that a dependency is not needed,
+or that a missing package is not listed on the package requirements.
+
+Fortunately, `z3c.dependencychecker` also has a solution for it.
+
+Add a `pyproject.toml` file on the root of your project with the following content:
+
+    [tool.dependencychecker]
+    ignore-packages = ['one-package', 'another.package' ]
+
+`z3c.dependencychecker` will totally ignore those packages in its reports,
+whether they're requirements that appear to be unused,
+or requirements that appear to be missing.
+
 Credits
 -------
 
--- a/z3c/dependencychecker/tests/test_report.py
+++ b/z3c/dependencychecker/tests/test_report.py
@@ -94,6 +94,35 @@ def test_missing_requirements_with_user_mapping(capsys, minimal_structure):
     assert 'Zope2' not in out
 
 
+def test_missing_requirements_with_ignored_packages(capsys, minimal_structure):
+    path, package_name = minimal_structure
+    write_source_file_at(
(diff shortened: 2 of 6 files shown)


README.rst
```diff
@@ -68,6 +68,27 @@ Add a `pyproject.toml` file on the root of your project with the following conte
 
 z3c.dependencychecker will read this information and use it on its reports.
 
+Ignore packages
+---------------
+
+Sometimes you need to add a package on `setup.py` although you are not importing it directly,
```
reinout: "on" setup.py => "in" setup.py

README.rst
```diff
@@ -68,6 +68,27 @@ Add a `pyproject.toml` file on the root of your project with the following conte
...
+or that a missing package is not listed on the package requirements.
+
+Fortunately, `z3c.dependencychecker` also has a solution for it.
+
+Add a `pyproject.toml` file on the root of your project with the following content:
+
+    [tool.dependencychecker]
+    ignore-packages = ['one-package', 'another.package' ]
+
+`z3c.dependencychecker` will read this information and use it on its reports.
```
reinout: `z3c.dependencychecker` will totally ignore those packages in its reports, whether they're requirements that appear to be unused or requirements that appear to be missing.

z3c/dependencychecker/tests/test_report.py
```diff
@@ -225,6 +290,37 @@ def test_unneeded_requirements(capsys, minimal_structure):
     assert 'two' in out
 
 
+def test_unneeded_requirements_with_ignored_packages(
+    capsys,
+    minimal_structure
+):
```
reinout: This `):` is in a weird location. It is apparently legal, but I guess it looks better if you place it on the line above.

gforcada: I really like to keep lines clean and not put everything and the kitchen sink there. Specially for future changes, if we need to add another parameter, a diff adding a single line would be enough, if we keep the parenthesis on it, we will have to change two lines.

But as it is a test and I want this to be merged and released I'm changing it :wink:

README.rst
gforcada: Changed

README.rst
gforcada: Oops :blush:

gforcada:
Actually I had the changelog entry written, I somehow forgot to commit it! :sweat_smile:

z3c/dependencychecker/tests/test_report.py
gforcada: I had to indent it *twice* (i.e 8 spaces) if not code analysis was complaining :)

gforcada:
@reinout thanks for all the feedback! I implemented all the suggestions, please have a second look and let me know if you would like to make further changes.

z3c/dependencychecker/tests/test_report.py
reinout: Ah, the "indentation continued on the same line" or so message? That is about the only pep8 warning that I dislike.

reinout: Hm, there's something weird. I see that you've changed it, but I don't see the commit. And if I click on the original commit github can't find anything and gives a message "this sometimes happens with a forced push".

Did you do anything creative with git? :-)

There are other functions in that testfile with the same indentation style: leave them be. It might just be something I've got to get used to.

PR #83, opened 2018-06-19
Exit code

gforcada:
Finally! :tada: with this one, the endeavor that started with https://github.com/reinout/z3c.dependencychecker/commit/465a3131abaf016930a78245a722394c2f30b75c *ends* now :smile: 

@reinout thanks for all your patience and wisdom!! :bowing_man:

### DIFF
--- a/z3c/dependencychecker/main.py
+++ b/z3c/dependencychecker/main.py
@@ -22,6 +22,11 @@ def main():
     report = Report(package_analyzed)
     report.print_report()
 
+    if options.exit_status:
+        exit(report.exit_status)
+
+    exit(0)
+
 
 def parse_command_line():
     usage = (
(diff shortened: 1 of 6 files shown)


z3c/dependencychecker/main.py
```diff
@@ -38,6 +43,13 @@ def parse_command_line():
         default=False,
         help='Show debug output',
     )
+    parser.add_option(
+        '--exit-zero',
+        action='store_false',
+        dest='exit_status',
+        default=True,
```
reinout: Having a positive variable with a default of True always helps a lot when reasoning about it. So: good!

(Stuff like `if omit_statuscode == False` always makes my head hurt.)

z3c/dependencychecker/main.py
```diff
@@ -22,6 +22,11 @@ def main():
     report = Report(package_analyzed)
     report.print_report()
 
+    if options.exit_status:
```
reinout: Both `options.exit_status` and `report.exit_status` have the same name.
It might be clearer to rename the option variable to `exit_with_statuscode`?

But.... It is only used in one place, so leave it be :-)

reinout:
Thank *you* for doing all the actual work :-)

I've released it as 2.3.

PR #86, opened 2018-06-30
Fix rendering of long description

gforcada:
I'm actually not even sure that it will fix it, only an actual release, a pre-release one?, would actually verify that.
But comparing our setup.py with other projects' setup.py I noticed that the ``keywords`` argument is a string rather than a list, so maybe changing that back fixes it.

Actually if you look at the pypi page of z3c.dependencychecker you see that at the very beginning it has some data that is not even on the README... I guess it should be on the left column but, maybe, the keywords-as-a-list breaks its rendering...

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@ Changelog of z3c.dependencychecker
 2.5 (unreleased)
 ----------------
 
-- Nothing changed yet.
-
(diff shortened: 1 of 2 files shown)


reinout:
Didn't help yet.

reinout:
I've done some other checks, too, but everything looks good.

PR #87, opened 2018-07-04
setup.py changes for long description

reinout:
Follow-up to #86 

- Copied zest.releaser's `read()` method with encoding handling. Perhaps it helps.

- Made 'description' into a single line instead of multi-line.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changelog of z3c.dependencychecker
 2.5 (unreleased)
 ----------------
 
-- Nothing changed yet.
+- Another go at fixing the long description in pypi.
+  [reinout]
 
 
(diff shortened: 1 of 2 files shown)


reinout:
I wanted to experiment with zest.releaser's keeping-encodings-happy setup.py trick, perhaps that could work.

Then I noticed that the `description` (not the `long_description`) contained newline characters. I'm not sure that's allowed, so I changed it to a single line.

Let's hope one of them works :-)

gforcada:
Let's see...

reinout:
I'll release it.

reinout:
Yeah! It worked. My guess is that it was the description with the newlines.

reinout:
See https://reinout.vanrees.org/weblog/2018/07/04/pypi-description.html

PR #89, opened 2018-07-05
[PackageMetadata] Check more locations

gforcada:
The world is vast, and the top level folder and ``src`` is not enough.

Specially in projects like Plone where quite a lot of packages have
``plone`` as their package_dir.

This commit makes ``package_dir`` search to look at all top level
folders.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@ Changelog of z3c.dependencychecker
 2.5 (unreleased)
 ----------------
 
-- Nothing changed yet.
-
+- Check in every top level folder if the .egg-info folder is in them.
+  [gforcada]
 
 2.4.4 (2018-07-04)
 ------------------
--- a/z3c/dependencychecker/package.py
+++ b/z3c/dependencychecker/package.py
@@ -47,10 +47,18 @@ class PackageMetadata(object):
     @cached_property
     def package_dir(self):
         """Check where the .egg-info is located"""
-        try_paths = (
-            self.distribution_root,
-            os.path.join(self.distribution_root, 'src', )
-        )
+        try_paths = [
+            self.distribution_root
+        ]
+        top_level_elements = [
+            os.path.join(self.distribution_root, folder)
+            for folder in os.listdir(self.distribution_root)
+        ]
+        try_paths += [
+            folder
+            for folder in top_level_elements
+            if os.path.isdir(folder)
+        ]
         for path in try_paths:
             folder_found = self._find_egg_info_in_folder(path)
             if folder_found:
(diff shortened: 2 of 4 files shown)


z3c/dependencychecker/package.py
```diff
@@ -47,10 +47,14 @@ def setup_py_path(self):
...
-            self.distribution_root,
-            os.path.join(self.distribution_root, 'src', )
-        )
+        try_paths = [
+            self.distribution_root
+        ]
+        try_paths += [
+            '{0}/{1}'.format(self.distribution_root, folder)
+            for folder in os.listdir(self.distribution_root)
+            if os.path.isdir('{0}/{1}'.format(self.distribution_root, folder))
```
reinout: I generally want to do the os.path.join just once:

    toplevel_items = [...join... for item in os.listdir(...)]
    try_paths += [folder for folder in toplevel_items if os.path.isdir(folder)]

That way you don't have a duplicate in the "if". But that's just a style thingy. Having just one list comprehension also is ok.

reinout:
The coverage went down a little [https://coveralls.io/builds/17855950/source?filename=z3c%2Fdependencychecker%2Fpackage.py#L73](link).

The new code you added uses os.path.isdir(), so the `if os.path.exists()` check [https://github.com/reinout/z3c.dependencychecker/blob/f4d69c4f017081477f07e48f980c77719c9bd16d/z3c/dependencychecker/package.py#L72](here) can be removed, I think. It was previously needed because we passed `src/` in without checking if it existed.

So: the new code works very well :-)

gforcada:
@reinout thanks for the review, I applied all the suggestions :-)

CHANGES.rst
```diff
@@ -4,8 +4,8 @@ Changelog of z3c.dependencychecker
 2.5 (unreleased)
 ----------------
 
-- Nothing changed yet.
-
+- Check in every top level folder if the .egg-info folder is in them.
```
reinout: I think there's a phrase at the top of `USAGE.rst` that talks about "egg-info in the root or in src/", that one also has to be adjusted.

reinout: @gforcada, I've fixed up the USAGE.rst documentation a bit in https://github.com/reinout/z3c.dependencychecker/pull/89/commits/938c4e6f21a380e47c99a6ae3f67a1aeb9d11e54 (I pushed it to your branch).

Does that look OK?

If so: can *you* do the merge + release? Then we know for sure that I set the right permissions on the pypi project :-)

gforcada: Thanks for the fix!

I will do the release, yes, thanks for granting me rights!

PR #90, opened 2018-07-08
User mappings everywhere

gforcada:
I was too lazy when I added user mappings, that I only added it to the first two reports (missing dependencies and missing test dependencies).

But while trying to fix a package in Plone I was already in the need of using user mappings for the other three reports.

So, finally, here it is.


reinout:
I released it as 2.6.

reinout:
(The other PR is also in).

PR #91, opened 2018-07-08
[modules] Consider docstrings imports as tests

gforcada:
Any import in a docstring is not used when running the code in a normal
mode, but rather only when called with some special tools, like a test
runner.
    
Thus, always consider the imports done there to be test imports.


reinout:
Looks good.

PR #92, opened 2018-07-21
Fix report

gforcada:
Test and fix the 'requirement should be test requirement' report to
account for imports of a user mappings used in both regular and test
imports that the report wrongly considered that the user mapping should
be downgraded to a test import only.

That was a hard nut to crack :sweat_smile: Thanks to the Plone PLIP https://github.com/plone/Products.CMFPlone/issues/2448 I guess we will find, hopefully not so many, some more bugs.

Fortunately, as the test suite is quite good, specially the ``test_report.py``, we could eventually, refactor the reports to make them readable, with a high degree of confidence.

### DIFF
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -146,12 +146,20 @@ class ImportsDatabase(object):
             self.imports_used,
             non_testing_filters,
         )
+        complete_non_testing_imports = []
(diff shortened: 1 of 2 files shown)


reinout:
Looks fine! I'll merge it now that I'm back from vacation :-)

reinout:
I released it as 2.7

PR #94, opened 2022-10-25
Use GitHub actions

gforcada:
As well as pip-tools to keep the dependencies up-to-date and a slew of linters and formatters to ensure the code stays fine and sound ✨


reinout:
Hi! What an amount of work! Thanks.
I'll have to take a bit more time later today to look it over a bit better, but the first impression is "hurray".
Replacing buildout with pip-tools is something that is my preference, too. isort and black: the same.

One thing surprised me right away: I'm totally used to black's `"double quoted strings"` and think they're a good idea. But the new code has single quotes. Huh? I didn't spot the config file that arranges that yet. Or did black change its default?

gforcada:
thanks! 😄 sorry that I forgot to mention what's TODO/draft 😓 

As for the quotes, I very much prefer single ones, but I don't mind changing them to double, I think the _configuration_ is using `flake8-quotes` ([pypi](https://pypi.org/project/flake8-quotes))

There are a couple of tests that fail and are skipped right now.

And there is a major problem 💥 we depend on [stdlib-list](https://pypi.org/project/stdlib-list) but it only supports up to python 3.9.

It might not be too bad though 🤔

reinout:
Ouch, I've left this lying around for far too long, sorry.
Looks good, I'm going to merge it.

One thing that needs handling separately: the quotes. I see you're passing `--skip-string-normalization` to the black check in the github action. That prevents black from complaining about the double quotes. So, if we want to keep the behaviour, it should be put into the `setup.cfg` as a standard option for black, otherwise I'm going to ruin the formatting upon every call to "black". I assume you've set it somewhere in a `~/.black` config file?

What I don't get: the option keeps black from *changing* the string character. But: how do you make sure all the double quotes are automatically changed to single ones?

The code looks really, really strange to me. I've gotten completely used to the standard `black` output and I kinda enjoy the fact that there doesn't have to be any discussion about the formatting anymore :-)

gforcada:
oh 😱 I also forgot about it and did not fix the failing tests, only made them be skipped 😕 let me try to fix them!

PR #96, opened 2022-11-29
Python 3.10 3.11 support

gforcada:
It is built on top of #95 so let's merge that first :-)


z3c/dependencychecker/db.py
```diff
@@ -1,13 +1,18 @@
...
 from z3c.dependencychecker.dotted_name import DottedName
 
 logger = logging.getLogger(__name__)
 
 
+# starting from python 3.10 the list of builtin methods are available directly
+PY_10_OR_HIGHER = True
+if sys.version_info[1] < 10:
+    PY_10_OR_HIGHER = False
+    from stdlib_list import stdlib_list
```
reinout, line 13: The 10-or-higher constant and the stdlib_list import are related/intermingled. I'd rather just have the constant here:

`PY_10_OR_HIGHER = (sys.version_info[1] >= 10)`

And then later on in the `_build_std_library()` you can do the import inside the `else`. That way you also don't have a global import that isn't imported part of the time. Seems like a nice way to dis-entangle the two.

reinout:
The other PR is merged, so you can fetch and merge the new master into this PR.

z3c/dependencychecker/db.py
gforcada: Good idea, I fixed it 👍🏾

gforcada:
It should be ready for review now 🍀

reinout (approved):

PR #99, opened 2023-01-17
Ignore Zope dependency

### LINKED ISSUE #98: Package hiding others
Not sure how this can actually be solved:

We have plenty of packages that depend on `Zope2`, that is fine, and as for `import` statements and such, with the mappings, all is good.

But since `Zope` is no longer on version 2 but rather on 5, the package got renamed to `Zope`.

Now if a package depends on `zope.schema` and `zope.interface` but also on `Zope` itself, `z3c.dependencychecker` fails to recognize that `zope.interface` and `zope.schema` are missing from `setup.py` since the package depends on `Zope` 💥 😱 

Since in this case [Zope](https://github.com/zopefoundation/Zope/tree/master/src) only is an umbrella and has no direct code, could we single it out in `z3c.dependencychecker`? 🍀 

And whenever another package does the same (let's say `Plone` 😅 ) we could add a key on `pyproject.toml` configuration that allows to list umbrella packages that provide no code on their own namespace? 🤔 

Ideas?

gforcada:
Fixes #98

### DIFF
--- a/z3c/dependencychecker/package.py
+++ b/z3c/dependencychecker/package.py
@@ -121,6 +121,9 @@ class PackageMetadata:
         this_package = self._get_ourselves_from_working_set()
         requirements = this_package.requires()
         for requirement in requirements:
+            if requirement.project_name == "Zope":
+                logger.debug("Ignoring 'Zope' requirement")
+                continue
             yield DottedName.from_requirement(
                 requirement,
                 file_path=self.setup_py_path,
@@ -135,6 +138,7 @@ class PackageMetadata:
             dotted_names = (
                 DottedName.from_requirement(req, file_path=self.setup_py_path)
                 for req in extra_requirements
+                if req.project_name != "Zope"
             )
             yield extra_name, dotted_names
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@ Changelog of z3c.dependencychecker
 2.9 (unreleased)
 ----------------
 
-- Nothing changed yet.
-
+- Ignore `Zope` package, as otherwise it swallows all `zope.` namespace packages.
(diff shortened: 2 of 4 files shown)


z3c/dependencychecker/package.py
```diff
@@ -121,6 +121,8 @@ def get_required_dependencies(self):
         this_package = self._get_ourselves_from_working_set()
         requirements = this_package.requires()
         for requirement in requirements:
+            if requirement.project_name == "Zope":
+                continue
```
reinout, line 125: Would a `logger.debug("Ignoring 'Zope' requirement")` or so be a good idea?

reinout (approved):
One logging suggestion, but I leave it to you to see how important it is to add it :-)

z3c/dependencychecker/package.py
gforcada: Should we do the same on the extras requirements? As we do it within a list comprehension it is a bit trickier to do it though :-/

reinout: Ouch, yes, that's trickier.
Well, I guess Zope will normally be a normal requirement, not somewhere in the extras? Fine enough if the main usecase is being logged.

gforcada: Yes, that's what I though as well, but wanted to double check with you 😄

gforcada:
@reinout could we get this merged and released? 😄 🤞🏾 🍀

reinout:
`2.9` is out now.

I'm totally fine with you merging&releasing finished branches, btw. I think you've got enough rights on pypi for that?

gforcada:
Yes, I think so,but as it'a been a while I wanted to ask confirmation first 😃

And thanks for the release 👍

PR #102, opened 2023-02-19
Ignore certain folders

### LINKED ISSUE #65: Speed improvement
z3c.dependencychecker is not slow per se,  but both implementations have a shared problem:  they keep scanning the filesystem for files with possible imports.

On the new implementation for example that happens for every single class in modules.py.

I see two different approaches on optimizing this:

- use VCS information (we should add another warning about it then)
- use MANIFEST.in information (again a new warning would be useful)

The big downside would be testing, either approach would make tests a bit more complex, though with a pytest fixture the solution can be elegant enough to not be too cumbersome.

Ideas? Should that be a real concern? 🤔

gforcada:
Partially fixes #65 

I ignored both `node_modules` and `__pycache__`, more folder names can be easily added.

For the python scanner we are already removing all folders that do __not__ contain an `__init__.py` file, which might bite us eventually 🤔 as some recent python/setuptools versions allow implicit namespaces without an `__init__.py`, I can't remember the PEP number, I can look it up, but that's beyond this PR 😄

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,14 +4,15 @@ Changelog of z3c.dependencychecker
 2.11 (unreleased)
 -----------------
 
-- Nothing changed yet.
-
+- Ignore `node_modules` and `__pycache__` folders
+  when scanning for files with dependencies.
+  [gforcada]
 
 2.10 (2023-01-30)
 -----------------
 
 - Do not ignore `Zope` user mappings, fixes previous release.
-  [gforcada
+  [gforcada]
 
 2.9 (2023-01-23)
 ----------------
--- a/z3c/dependencychecker/modules.py
+++ b/z3c/dependencychecker/modules.py
@@ -27,6 +27,8 @@ TEST_IN_PATH_REGEX = re.compile(TEST_REGEX, re.VERBOSE)
 
 logger = logging.getLogger(__name__)
 
+FOLDERS_TO_IGNORE = ("node_modules", "__pycache__")
+
 
 class BaseModule:
     def __init__(self, package_path, full_path):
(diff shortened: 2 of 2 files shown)


reinout (approved):
Looks good!

mauritsvanrees:
> For the python scanner we are already removing all folders that do **not** contain an `__init__.py` file, which might bite us eventually 🤔 as some recent python/setuptools versions allow implicit namespaces without an `__init__.py`, I can't remember the PEP number, I can look it up, but that's beyond this PR 😄

That is PEP-420 with native namespaces, and is available since Python 3.3 already.  Best documentation for the three namespace alternatives is here:
https://packaging.python.org/en/latest/guides/packaging-namespace-packages/

Problem with switching to native/implicit namespaces from the viewpoint of Plone is that you would need to switch *all* packages that use the same namespace all at once.  Mostly: all `plone.*` and `plone.app.*` packages.  And if we want to switch the `Products.*` packages as well, we would need to coordinate with Zope, as that contains `Products.*` packages as well.  Same for all `collective.*` packages for add-ons, otherwise you would only be able to install a part of these.  It can be done, but is a lot of work, and I am not sure what we would gain.
But I do think recently I started seeing warnings that non-native namespaces might be deprecated, which could mean support will be dropped at some point in the future, depending on your Python/pip/setuptools version.  But I hope I have misread this.

Anyway, this PR is not the place for such a discussion, just wanted to share my thoughts. :-)
But if there is a Python scanner here that ignores a directory **and its sub directories** when no `__init__.py` file is there, then this may need to change.

gforcada:
Yes, we talked about it on Innsbruck 😄 and I hope that you misread that indeed, otherwise having to do a complete AND synchronous change of all plone packages and add-ons might be a way too much 😅 

For z3c.dependencychecker, yes, we might have to support that, but given that is not so popular(?) and there are not that many requests, we might get away with it as it stands for a bit more.

Otherwise relying on VCS information might be much better 👍🏾

PR #101, opened 2023-01-29
Do not ignore zope user mapping

gforcada:
Closes #100

PR #113, opened 2023-04-16
Fix zope user mapping

### LINKED ISSUE #112: More Zope mapping problems
With #98 and #100 it sort of feel like we were fixing the problem but turns out there's more 😓 

See the following scenario:

- a package has a user mapping `Zope = ['foo']`
- does __NOT__ declare `Zope` as a dependency
- has any random `zope.interface` import

One would expect to get `zope.interface` as undeclared dependency, right?

Turns out that right now, it is not 😕 

On #99 we always were taking the `Zope` user mapping, and in a way, assuming that the key (`Zope`) is an actual dependency, but not checking for it actually.

For regular scenarios where Zope is actually a dependency, that works great, but if it is not, then it hides all dependencies 💥 

Reverting #99 changes makes it work again, I will double check all other use cases and improve the tests 🚧

gforcada:
Hopefully closes #112 

That was a hard nut to crack, but fortunately with tests, things are much easier to spot and reason about 🍀 

More or less both #99 and #101 changes are removed and the commit about the umbrella distributions fixes it properly. I added two test matrix to ensure we do not miss any case.

Hopefully the, plenty of, comments help to understand what is being tested 🤞🏾

### DIFF
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -339,7 +339,7 @@ class ImportsDatabase:
         # called some lines below
         iterable = (item for item in obj_list if obj in item)
 
-        # any builtin stops consuming the iterator as soon as an element of the
+        # `any` stops consuming the iterator as soon as an element of the
         # iterable is True
         if any(iterable):
             # if 'obj' is found, the pipeline needs to stop and discard the
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -73,12 +73,13 @@ class ImportsDatabase:
 
         if package.name != "Zope" and package not in self._all_requirements():
             logger.info(
-                "Ignoring package %s as is not a dependency of the "
+                "Ignoring user mapping %s as is not a dependency of the "
                 "package being analyzed",
                 package,
             )
             return
 
+        logger.debug("Using user mapping %s", package)
         self.user_mappings[package] = set(packages_provided)
 
         for single_package in packages_provided:
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -59,13 +59,13 @@ class ImportsDatabase:
             self._filter_out_python_standard_library,
         )
         for single_import in imports:
-            logger.debug("    Import found: %s", single_import.name)
 
             unknown_import = self._apply_filters([single_import], filters)
             if unknown_import:
+                logger.debug("    Import found: %s", single_import.name)
                 self.imports_used.append(single_import)
             else:
-                logger.debug("    Import ignored: %s", single_import.name)
+                logger.debug("    Import found & ignored: %s", single_import.name)
 
     def add_user_mapping(self, package_name, provided_names):
         package = DottedName(package_name)
--- a/z3c/dependencychecker/db.py
+++ b/z3c/dependencychecker/db.py
@@ -71,7 +71,7 @@ class ImportsDatabase:
         package = DottedName(package_name)
         packages_provided = [DottedName(name) for name in provided_names]
 
-        if package.name != "Zope" and package not in self._all_requirements():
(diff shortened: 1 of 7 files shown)


mauritsvanrees (approved):
I have not studied the code, but it seems to work.  I tried it out on `plone.app.event`.  With `z3c.dependencychecker` master it says no problem (well, `GenericSetup` missing, but let's ignore that here).  With the branch from this PR it complains about all kinds of missing dependencies, like `Acquisition` (part of `Zope` in our mapping) and `zope.annotation`.  When I add `Zope` to the requirements, it stops complaining.

I think we should add `Plone` to the umbrella distributions for the same reason.  The [`Plone`](https://github.com/plone/Plone/) package has no code of its own, it only pulls in dependencies.  When I remove all `plone.*` package from the `setup.py` of `plone.app.event`, and add `Plone`, `dependencychecker should complain, but it does not.  Adding it to the umbrella distributions, makes dependencycheker rightfully complain again.
(For an add-on it might be useful to have a user mapping from `Plone` to a few hundred other packages, but it would be up to the add-on to define this mapping.)

z3c/dependencychecker/db.py
```diff
@@ -59,26 +59,27 @@ def add_imports(self, imports):
...
             else:
-                logger.debug("    Import ignored: %s", single_import.name)
+                logger.debug("    Import found & ignored: %s", single_import.name)
 
     def add_user_mapping(self, package_name, provided_names):
         package = DottedName(package_name)
         packages_provided = [DottedName(name) for name in provided_names]
 
-        if package.name != "Zope" and package not in self._all_requirements():
+        if package not in self._all_requirements():
```
reinout, line 74: Nice. It always looks good if a hardcoded exception can be removed.

z3c/dependencychecker/dotted_name.py
```diff
@@ -3,6 +3,15 @@
 
 from cached_property import cached_property
 
+# An umbrella distribution is a Pypi distribution, like `Zope`,
+# that the importable packages that it provides have different namespaces,
+# that the umbrella distribution name.
```
reinout, line 8: Ehm, there's something wrong with the sentence. Something like this:

```suggestion
# An umbrella distribution is a Pypi distribution, like `Zope`,
# of which the importable packages that it provides have different namespaces
# than the umbrella distribution name.
```

z3c/dependencychecker/tests/test_report.py
```diff
@@ -627,3 +629,151 @@ def test_exit_status_set(minimal_structure):
     assert report.exit_status == 0
     report.print_report()
     assert report.exit_status == 1
+
+
+PKG_REQUIRES_LINE = ("requires.txt", "zope.interface")
+ZOPE_MAP_REQUIRES_LINE = ("requires.txt", "Zope")
+MAPPING_LINE = ("pyproject.toml", '[tool.dependencychecker]\nZope = ["zope.interface"]')
+CODE_LINE = ("__init__.py", "import zope.interface.Interface")
```
reinout, line 637: Nice way of testing!

I initially got a bit distracted by the term `_LINE`, but I think that can be fixed with a comment:

```suggestion
# These 'lines' are instructions for the parametrized test below. A filename + the desired contents.
PKG_REQUIRES_LINE = ("requires.txt", "zope.interface")
ZOPE_MAP_REQUIRES_LINE = ("requires.txt", "Zope")
MAPPING_LINE = ("pyproject.toml", '[tool.dependencychecker]\nZope = ["zope.interface"]')
CODE_LINE = ("__init__.py", "import zope.interface.Interface")
```

reinout (approved):
Looks good to me, I've got two textual adjustment to comments.

For the actual zope/plone namespace stuff: you and @mauritsvanrees are the best positioned to say anything useful about that :-)

PR #114, opened 2023-04-18
ZCML `implements` directive

### LINKED ISSUE #110: Report `<implements interface` ZCML attributes?
While checking `plone.app.dexterity` with `z3c.dependencychecker` I noticed the following line:

```xml
    <implements interface="plone.contentrules.engine.interfaces.IRuleAssignable" />
```

Still `z3c.dependencychecker` was reportin that `plone.contentrules` is not needed. We probably don't parse and extract this `interface` attribute.

Should we? 🤔

gforcada:
Closes #110 

`pytest.mark.parametrize` is wonderful ✨ 🪄 🐰 

The first commit is where the actual `<implements` is added, the second is a big refactoring of the tests to use the same approach used for `<implements`.

We could go one step further and combine all tests into a single giant test, but maybe is too much? 😅 

Notice that if you add multiple `@pytest.mark.parametrize` it does the product of the elements, i.e.

```python
@pytest.mark.parametrize("one,two", (1,2), (3,4))
@pytest.mark.parametrize("three,four", (5,6), (7,8), (9,10))
```

Will run:
- `1,2`, `5,6`
- `1,2`, `7,8`
- `1,2`, `9,10`
- `3,4`, `5,6`
- `3,4`, `7,8`
- `3,4`, `9,10`

### DIFF
--- a/z3c/dependencychecker/tests/test_modules_zcml.py
+++ b/z3c/dependencychecker/tests/test_modules_zcml.py
@@ -1,5 +1,7 @@
 import os
 
+import pytest
+
 from z3c.dependencychecker.modules import ZCMLFile
 from z3c.dependencychecker.tests.utils import write_source_file_at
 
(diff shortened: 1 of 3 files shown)


reinout:
Thanks! I didn't know about the multiple-parametrize possibility. Handy. I've immediately told my colleagues about it :-)

z3c/dependencychecker/tests/test_modules_zcml.py
```diff
@@ -67,172 +72,69 @@ def test_create_from_files_deep_nested(minimal_structure):
...
+    _verify_dotted_names(dotted_names, imports, result)
+
+
+@pytest.mark.parametrize("attribute", ("for", "factory", "provides"))
+@pytest.mark.parametrize("result,imports", TEST_IMPORTS_MATRIX)
+def test_zcml_adapter(tmpdir, result, imports, attribute):
+    """Check that the ZCML <adapter directive is scanned."""
+    zcml_stanza = f'<adapter {attribute}="{imports}" />'
+    dotted_names = _get_zcml_imports_on_file(tmpdir, zcml_stanza)
+    _verify_dotted_names(dotted_names, imports, result)
```
reinout, line 89: 2x parametrize decorator *and* f-strings using those attributes: fancy! Nice solution.

reinout (approved):
Note: I haven't looked at the test failures yet. Code looks good, though.

PR #116, opened 2023-04-18
Handle modern distributions

### LINKED ISSUE #107: Parse `pyproject.toml` and `setup.cfg` files as well
`setup.py` is being phased out generally in python projects favoring either `setup.cfg` to contain the metadata (and thus the dependencies), or even better `pyproject.toml`.

Currently `z3c.dependencychecker` refuses to work if no `setup.py` is found at all.

gforcada:
Closes #107 

Actually, we _never_ parse `setup.py`, but rather get the information from the `egg-info` folder. So we might skip completely any reference to the configuration files and ask only to build the distribution in a way that an `egg-info` is created.

### DIFF
--- a/z3c/dependencychecker/package.py
+++ b/z3c/dependencychecker/package.py
@@ -39,7 +39,7 @@ class PackageMetadata:
                 break
         else:
             logger.error(
-                "No distribution configuration file was not found in %s. "
+                "No distribution configuration file was found in %s. "
                 "Without it dependencychecker can not work.",
                 self._path,
             )
--- a/z3c/dependencychecker/package.py
+++ b/z3c/dependencychecker/package.py
@@ -39,7 +39,7 @@ class PackageMetadata:
                 break
         else:
             logger.error(
-                "setup.py was not found in %s. "
+                "No distribution configuration file was not found in %s. "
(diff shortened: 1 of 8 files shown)


z3c/dependencychecker/package.py
```diff
@@ -28,21 +34,26 @@ def __init__(self, path):
...
     @cached_property
     def distribution_root(self):
-        if "setup.py" not in os.listdir(self._path):
+        for metadata_file in METADATA_FILES:
+            if metadata_file in os.listdir(self._path):
+                break
+        else:
             logger.error(
-                "setup.py was not found in %s. "
+                "No distribution configuration file was not found in %s. "
```
mauritsvanrees, line 42:
```suggestion
                "No distribution configuration file was found in %s. "
```

mauritsvanrees (approved):
LGTM.
One inline comment about fixing a typo.

z3c/dependencychecker/package.py
reinout, line 42: "Distribution configuration file" sounds a bit abstract. `pyproject.toml/setup.py/.cfg` is more explicit (and shorter).

reinout (approved):

PR #117, opened 2023-04-18
feat: detect conditional imports

### LINKED ISSUE #106: Detect if imports are done within try/except blocks
It is quite common, for backwards compatibility code, to have fallback imports, i.e.

```python
try:
    from new.shiny import stuff
except ImportError:
    import old.hazard import stuff
```

It would be good if `z3c.dependencychecker` does not suggest that both `new.shiny` and `old.hazard` should be declared as dependencies.

Not sure exactly what it should suggest though?

- ❓ none
- ❓ only one, which one?
- ❌ both

None feels _wrong_, as either one or the other should be needed to have `stuff` working down the module.

Both feels _way wrong_, because having to imports providing the same from different packages is probably what you don't want.

Thinking a bit more, maybe mentioning them both and letting the user decide about it (i.e. not reporting it as broken) might be better, as the example above is simple, but it can be more complex, so a default option might not be desirable.

Opinions? 🍀

gforcada:
Closes #106 

This is a rough draft, I tested with `Products.CMFPlone` and at least the code does not break 😅 

Wording and code organization, as well as tests are missing/can be improved.

Do you think the direction is the right one? 🤔

### DIFF
--- a/z3c/dependencychecker/modules.py
+++ b/z3c/dependencychecker/modules.py
@@ -29,6 +29,7 @@ logger = logging.getLogger(__name__)
 
 FOLDERS_TO_IGNORE = ("node_modules", "__pycache__", "venv")
 
+HAS_REPORTED_CONDITIONAL_IMPORTS = False
 
 class BaseModule:
     def __init__(self, package_path, full_path):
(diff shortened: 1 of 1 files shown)


mauritsvanrees (commented):
I have not tried it, but the direction seems okay.  At least it is useful to report that at least one conditional import is found.

One inline question because I don't know the code.

z3c/dependencychecker/modules.py
```diff
@@ -115,11 +116,31 @@ def _process_ast_node(self, node):
...
+        elif isinstance(node, ast.Try):
+            if not HAS_REPORTED_CONDITIONAL_IMPORTS:
+                self._is_a_conditional_import(node)
 
     @staticmethod
     def _is_relative_import(import_node):
         return import_node.level > 0
 
+    def _is_a_conditional_import(self, node):
+        global HAS_REPORTED_CONDITIONAL_IMPORTS
```
reinout, line 128: Normally `CONSTANTS` are pretty constant and don't change. This one does, which surprised me initially.
A short comment next to the original constant would help (`# Gets set to True when xyz is found`).

z3c/dependencychecker/modules.py
```diff
@@ -29,6 +29,7 @@
 
 FOLDERS_TO_IGNORE = ("node_modules", "__pycache__")
 
+HAS_REPORTED_CONDITIONAL_IMPORTS = False
```
reinout, line 32: Why the "reported"? It might have to do with code you'll add later. For now, it sounds like `CONDITIONAL_IMPORTS_FOUND` could be clearer?

z3c/dependencychecker/modules.py
```diff
@@ -115,11 +116,31 @@ def _process_ast_node(self, node):
...
 
+    def _is_a_conditional_import(self, node):
+        global HAS_REPORTED_CONDITIONAL_IMPORTS
+        for handler in node.handlers:
+            type_info = getattr(handler, 'type', False)
+            if type_info:
+                exception_id = getattr(type_info, 'id', False)
+                if exception_id == 'ImportError':
+                    print('This distribution has conditional imports')
+                    HAS_REPORTED_CONDITIONAL_IMPORTS = True
```
reinout, line 135: The basic idea was to just report it, right? Just for information?

Is it easy to at a filename+linenumber at this point?

PR #123, opened 2023-08-02
Drop python 3.7 and add 3.12

gforcada:
I was about to make a new release, and noticed that GitHub has two security reports about our dependencies (in requirements.txt) bumping them to the minimum version is still fine and doable on 3.7, but if we want to take the time to also update all other dependencies, then we already have problems with some packages (`pip-tools` at least) that have already dropped 3.7 support.

This PR does the necessary changes to drop 3.7 here as well.

Should we make that a major release, or a feature one is enough? I'm in favor of the later, as otherwise every year, whenever we drop support for the python version that gets EOL we would have to make a new major version, which feels odd.

Meanwhile we can do a last 3.7 release with only bumping the packages that have security related problems, which anyway are not direct dependencies of `z3c.dependencychecker` but rather from our development tooling. 🍀


mauritsvanrees:
It is fine to drop 3.7 support, because it is end of life since end of June.

I would be inclined to make it a major version.  I think that is what we have mostly been doing with Plone packages, except for our two main packages `Plone` and `Products.CMFPlone`, which have their own considerations.  Same for Zope: the main `Zope` package is special and can drop Python support in a minor version, because you do not want a new major Zope version every year, but for `zope.*` packages a major release is done.

On the other hand, if you drop 3.7 support and declare this with `python_requires>=3.8`, then users of Python 3.7 will not even _find_ the new version when they `pip install` us, so it does not actually break anything...  So a minor version could be okay too.

reinout:
Minor version is okay with me.

gforcada:
@mauritsvanrees @reinout as it took me so much time to clean and fix this PR, 3.12 is already out, so I added it to our test matrix, and it does work out of the box 🎉 

Let me know if you would like other changes in, or otherwise we could merge and release 😄

reinout:
Merge+release is fine! You want to do that?

gforcada:
Done `2.13` is out 🎉 😄

PR #125, opened 2023-12-19
Pathlib everywhere

gforcada:
While looking at the code I noticed that we were using plenty of `os.sep` and `os.path.join` that since Python 3.4 is so much more nicer to use `pathlib.Path` instead.

The diff turned out quite big, but it's only about replacing old with new usages, no functionality changes.

As an extra bonus there are a couple of minor improvements added 😄

### DIFF
--- a/z3c/dependencychecker/tests/conftest.py
+++ b/z3c/dependencychecker/tests/conftest.py
@@ -11,6 +11,7 @@ import random
 import shutil
 import string
 import tempfile
+from pathlib import Path
 
 import pkg_resources
 import pytest
@@ -21,12 +22,12 @@ from z3c.dependencychecker.package import ImportsDatabase
 
 @pytest.fixture
 def fake_project():
-    temp_folder = tempfile.mkdtemp(prefix="depcheck")
+    temp_folder = Path(tempfile.mkdtemp(prefix="depcheck"))
     fake_package_files = pkg_resources.resource_filename(
         "z3c.dependencychecker.tests",
         "sample1",
     )
-    package_folder = os.path.join(temp_folder, "sample1")
+    package_folder = temp_folder / "sample1"
     shutil.copytree(fake_package_files, package_folder)
     # To prevent the sample .py files to be picked up by ourselves or other
     # tools, I'm postfixing them with ``_in``, now we get to rename them.
@@ -39,8 +40,8 @@ def fake_project():
             if new_filename == "_it__.py":
                 # Oopsie :-) The replace works too well...
                 new_filename = "__init__.py"
-            source = os.path.join(dirpath, filename)
-            target = os.path.join(dirpath, new_filename)
+            source = Path(dirpath) / filename
+            target = Path(dirpath) / new_filename
             os.rename(source, target)
 
     yield package_folder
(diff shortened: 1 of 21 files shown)


gforcada:
Interestingly, on all python versions works but on 3.10 🤷🏾

z3c/dependencychecker/tests/conftest.py
```diff
@@ -86,31 +83,21 @@ def _add_egg_info(folder):
...
 def _write_pkg_info_file(folder):
-    with open(os.path.join(folder, "PKG-INFO"), "w") as pkg_info:
-        lines = "\n".join(
-            ["Metadata-Version: 1.0", "Name: testpackage", "Version: 1.0.dev0"]
-        )
-        pkg_info.write(lines)
+    lines = "\n".join(
+        ["Metadata-Version: 1.0", "Name: testpackage", "Version: 1.0.dev0"]
+    )
+    (folder / "PKG-INFO").write_text(lines)
```
reinout, line 89: Not something that needs changing here, just a generic observation on my own habits.
My initial thought when I saw `.write_text(lines)` was "hey, can I put a list of strings into `.write_text()`? Reason is that I *nowadays* normally use "text" or "contents" for a multi-line string. `lines`, for me, is always a list of individual lines (=strings).

Now I'm wondering if I originally put in that `lines` years and years ago and whether I changed my habits :-)

z3c/dependencychecker/tests/conftest.py
```diff
@@ -86,31 +83,21 @@ def _add_egg_info(folder):
...
+    (folder / "requires.txt").write_text("\n".join(["one", "two"]))
 
 
-def _write_top_level_file(folder_path, package_name):
-    file_path = os.path.join(folder_path, "top_level.txt")
-    with open(file_path, "w") as top_level_file:
-        lines = "\n".join([package_name])
-        top_level_file.write(lines)
+def _write_top_level_file(folder, package_name):
+    (folder / "top_level.txt").write_text("\n".join([package_name]))
```
reinout, line 97: Nice: much shorter.

z3c/dependencychecker/tests/test_user_mappings.py
```diff
@@ -36,18 +34,11 @@
...
 def _write_user_config(path, content):
-    file_path = os.sep.join(
-        [
-            path,
-            "pyproject.toml",
-        ]
-    )
-    with open(file_path, "w") as config_file:
-        config_file.write(content)
+    (path / "pyproject.toml").write_text(content)
```
reinout, line 37: From 8 lines to 1: nice.

reinout (approved):

reinout:
The 3.10 error is caused by https://github.com/python/cpython/issues/88227

z3c/dependencychecker/tests/conftest.py
gforcada: I think, I rewrote the tests to use pytest long ago, so it might be my habits here 😄

reinout:
3.10 is still a problem. Weird. Looks like the bug I mentioned. 
Barbaric idea: remove 3.10 from the test? We're testing 3.9 and 3.11, if those two work...

gforcada:
Yes, we could do that indeed, maybe with a big disclaimer, or directly a `sys.exit(1)` if the user tries to run it on Python 3.10? 🤔

reinout:
I've removed 3.10 from the tests and added a note to the changelog. I assume most people using this tool use a more recent python version. And otherwise they'll probably look at the changelog when they see a sudden error.

reinout:
@gforcada , I did some mocking in [582fd4f](https://github.com/reinout/z3c.dependencychecker/pull/125/commits/582fd4fa10cf59574438b7915f7e91d012e6ea71), can you check that?

Weird, I must have done something wrong, as the code isn't reached?? Anyway, I'm off home to dinner :-)

reinout:
Coverage went down `0.001%` because we removed a few lines somewhere. Failing on that is too restrictive to my taste, so I've set the threshold for going down to a generous `0.5%` :-)

@gforcada, if you're happy with my mucking-about with the tests, you can merge it.

PR #126, opened 2023-12-20
Support implicit namespaces

### LINKED ISSUE #120: Support native namespace packages
@gforcada: We should support native namespace packages or disable this checking on plone/meta:

Ref: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages

gforcada:
Closes #120 

Seems that all the discussion we had with @ericof at the issue is not so relevant, as we are already using the `.egg-info` information to know where the source code is located 🎉

### DIFF
--- a/z3c/dependencychecker/modules.py
+++ b/z3c/dependencychecker/modules.py
@@ -27,7 +27,7 @@ TEST_IN_PATH_REGEX = re.compile(TEST_REGEX, re.VERBOSE)
 
 logger = logging.getLogger(__name__)
 
-FOLDERS_TO_IGNORE = ("node_modules", "__pycache__")
+FOLDERS_TO_IGNORE = ("node_modules", "__pycache__", "venv", ".venv")
 
 
 class BaseModule:
--- a/z3c/dependencychecker/modules.py
+++ b/z3c/dependencychecker/modules.py
@@ -69,12 +69,7 @@ class PythonModule(BaseModule):
             return
 
         for path, folders, filenames in os.walk(top_dir):
-            if "__init__.py" not in filenames:
-                # Don't descend further into the tree.
-                # Clear folders variable in-place.
-                folders[:] = []
-                continue
-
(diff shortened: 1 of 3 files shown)


z3c/dependencychecker/modules.py
```diff
@@ -27,7 +27,7 @@
 
 logger = logging.getLogger(__name__)
 
-FOLDERS_TO_IGNORE = ("node_modules", "__pycache__")
+FOLDERS_TO_IGNORE = ("node_modules", "__pycache__", "venv", ".venv")
```
reinout, line 30: Handy!

z3c/dependencychecker/modules.py
```diff
@@ -69,12 +69,7 @@ def create_from_files(cls, top_dir):
             return
 
         for path, folders, filenames in os.walk(top_dir):
-            if "__init__.py" not in filenames:
-                # Don't descend further into the tree.
-                # Clear folders variable in-place.
-                folders[:] = []
-                continue
-
+            folders[:] = [d for d in folders if d not in FOLDERS_TO_IGNORE]
```
reinout, line 72: The `__init__.py` rule probably ignored quite some folders. Should we now start to ignore all `.*` folders, too, for instance?

reinout: On the other hand.... that's something we'll discover in practice and that we can adjust to. I think the implementation is fine.

reinout (approved):
Looks handy!

gforcada:
May I merge and release then? 😄 I'm never comfortable doing that without asking and getting approval 😅 

But given that plenty of packages are configured with `plone/meta` which includes in one of its tox.ini/GHA/GitLab CI a check with `z3c.dependencychecker`, we need to get this PR released, otherwise we block any migration to implicit namespaces ✨

reinout:
Release is fine. 
Reinout

gforcada:
`z3c.dependencychecker==2.14` is out! 🎉

PR #128, opened 2023-12-28
Ignore all dot leading folders

gforcada:
@reinout indeed, ignoring all dot leading folders is important!!

I forgot to do some real testing, and once I did that, after the release, the first one already showed the problem: `.tox` stalls the processing 😕 

I refactored, a bit, the scanners to share the filtering logic on a single function 🧹

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@ Changelog of z3c.dependencychecker
 2.14.1 (unreleased)
 -------------------
 
-- Nothing changed yet.
-
+- Ignore any dot leading folder (like `.tox` or `.git`) while scanning for files.
+  [gforcada]
 
 2.14 (2023-12-28)
 -----------------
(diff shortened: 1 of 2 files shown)


gforcada:
We might want to expand this by allowing users to specify a list of folders to be ignored, probably.

`node_modules` is the default folder if you have some JS code within a python distribution, but the name might as well be another...

gforcada:
I did merge and release `z3c.dependencychecker==2.14.1` to fix this obvious performance regression 😓

reinout:
Common problem. `.gitignore`, `.dockerignore`... There are multiple places where I occasionally trip over to-ignore or to-include stuff. I wonder if there's some existing thing we can use? Probably we'll need a custom one.

gforcada:
Another approach would be to use something like `git ls-files` and only scan those files...

PR #129, opened 2023-12-28
Ignore folders in more code paths

gforcada:
The fix from #128 missed a few classes 😕


gforcada:
And `2.14.2` is out 🎉

PR #134, opened 2024-02-02
Trying out plone/meta

reinout:
@mauritsvanrees told me about plone/meta last weekend. Something like that is just what I need for my work projects :-)
So I'm testing it out with z3c.dependencychecker now.

There are two hacks that I need:

- ~An `ignore-bad-ideas` setting for check-manifests, but there's no way to add that via `.meta.toml`. Well, partially. Via extra_lines I can add it to the end of the pyproject.toml and purely by luck check-manifest's section is the last one.~ Fixed in https://github.com/plone/meta/pull/212 
- **Manually**, ~I removed `plone-test` (2x) from `tox.ini` as the tests won't run otherwise.~ Fixed in https://github.com/plone/meta/pull/212

TODO: fix up the documentation if we're really going to use this.

~Problem with plone/meta: the generated `.editorconfig` has a comment at the end of the line of two `[...]` sections. My emacs editorconfig-mode plugin doesn't like that. Probably something that needs fixing in the emacs plugin. **Update:** no, probably it needs fixing in meta.~ Fixed in https://github.com/plone/meta/pull/208

One thing surprised me: I wanted to use triple single quotes to quote the `extra_lines` value as it contains double quotes. But plone/meta converts the `.meta.toml` back to triple double quotes.


reinout:
I see the coverage ends up at 0%, so there's probably still some old config lying around.
And I don't know if it is a good idea to include this project in plone/meta as it is no real plone product.
Anyway, @gforcada will probably enjoy the fact that I'm trying this out :-) Feedback/fixes/updates welcome.

reinout:
Oh, and `tox -e circular` complains about `graphviz/cgraph.h` not being found. Perhaps an OSX thingy in combination with python 3.12?

gforcada:
Nice! 😊 🎉 

Please, propose more options if you need them, as you say for the `check-manifest` configuration 👍🏾 

The coverage report indeed, is not part of `plone/meta` but some `coveralls` integration that might be broken... not sure if all tests do pass then it might work again 🤷🏾 

Finally, I see that `circular` does work now but `dependencies` fails with a `stdlib_list`, which, I think, is a conditional import (should be fixed if we merge at some point #117 ). For now, we can add it on the `dependencies_ignores` configure option ✨

gforcada:
I see that the `meta.yml` related jobs coming from `plone/meta` fail due to missing some environment variables in GitHub (that should be fixed with https://github.com/plone/meta/pull/211).

For the local `test.yml` failures, we should stop running `flake8`, `isort` and such, as that's part of the `meta.yml` provided ✨ 

@reinout I took the liberty to add a couple of commits on that direction, hopefully you ok with them, otherwise feel free to revert them 👍🏾

reinout:
I wonder whether my `.coveragerc` is still needed with this setup?

gforcada:
We can try to remove it and see what the coverage says 😄 

At least, given that we would use tox, we can cleanup the list quite a bit I guess 🍀

reinout:
TODO (probably for myself) before I'm merging it: update the documentation ("tox -e xyz" instead of "venv/bin/pytest" and so).

We shouldn't put it into plone-meta's actual list of packages just yet because of the plone-test issue and the ignore-bad-ideas that I want to integrate a little bit better.

gforcada:
The list you refer to is a for a feature of `zopefoundation/meta` that allows you to mass update repositories that are listed there.

`plone/meta` did not bother to remove that feature, though it is not so useful for us, as we are moving the tool a bit away from being project centric (i.e. only for `zopefoundation` related repositories), but more of a Plone/python generic tool.

reinout:
I think I've fixed all my issues in https://github.com/plone/meta/pull/212

gforcada:
@reinout sorry! I was either too busy, or my brain was melting 😓 

Great job! 🌟

PR #152, opened 2026-02-27
Switch to native namespace (PEP 420)

gforcada:
Use native namespaces, that's yet another standard/best practice to follow, and it also is another way to get rid of `pkg_resources` 😄 

Built on top of #151 and #150 let's get those merged first

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Changelog of z3c.dependencychecker
 
 - Move distribution to src layout [gforcada]
 
+- Switch to native namespace (PEP 420) [gforcada]
+
 2.15 (2024-04-11)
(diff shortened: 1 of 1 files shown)


gforcada:
Seems that python 3.10 is somewhat broken for us, we were not testing it, @reinout would you be ok with skipping it and using 3.11 as the lowest python to be tested?

gforcada:
From 3.11 to 3.13 all seems good. 3.10 seems to be broken and for 3.14 there are some deprecations that need to be taken care of.

If that's fine, I would say to merge this, finish the porting away from `pkg_resources` and then we can look at 3.14 👍🏾

reinout (approved):

PR #156, opened 2026-03-02
Drop pkg_resources

gforcada:
Still on draft as it is not finished, but it is mostly to present the way I thought of: rely on `wheel-inspect` to get a JSON of metadata from a wheel, and with that and the source code, check the dependencies as usual 🎉 

https://pypi.org/project/wheel-inspect/


reinout (approved):
Approach looks good. I like it that many lines have been replaced by the wheel-inspect package.

gforcada:
Code should be ready to test and review, I'm running it against all plone repositories to see how much it differs from the current released version 🤞🏾

